PR #15, opened 2018-06-11
Managed btcpay

LinoxBE:
As discussed in Slack, adding the managed BTCPay servers to the docs.

### DIFF
--- a/Managed.md
+++ b/Managed.md
@@ -1,10 +1,10 @@
 # Managed BTCPay servers
 
-Are you looking for a managed BTCPay server?
+You are maybe interested in BTCPay Server, but you maybe don't have the necessary experience to set it up? Or you maybe want to be able to call somebody in case you have an issue? Please have a look at the following services, all have another focus:
 
 * [lightbo.lt](https://lightbo.lt)
 * [ltcpay.com](https://ltcpay.com/)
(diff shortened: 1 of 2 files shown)


NicolasDorier:
Can you would call it "Third party hosted BTCPay servers" and explain why a merchant would want it?

LinoxBE:
I'll change the naming, sounds good. But regarding hy a merchant would want it, quickly becomes marketing (in my eyes, at least). I'm not sure that you want that on such a page.

What makes more sense then maybe, is that every service/host adds 1 phrase behind his URL, highlighting why people should click there?

NicolasDorier:
This is not marketing. Hosting your own server is more complicated than just using a third party host. With third party host you know you have somebody to contact and help if something goes wrong.

LinoxBE:
Here we go. What do you think? Simple but straight-forward.

LinoxBE:
I forgot to change the title to "Third party hosted BTCPay servers". Should be good now.

NicolasDorier:
I am back to code, will be more reactive next time :p

LinoxBE:
No problem, it wasn't critical :-p

I hope that you enjoyed your time with your fresh wife, and your family ;-)

PR #17, opened 2018-06-24
Adding docs / video for Drupal Commerce 2.x

ndeet:
Also changed some http -> https


ndeet:
For all the other payment options (Paypal, Stripe, Square, etc.) you have Test/Sandbox URLs. With this option you can switch between Test and Live endpoints. 

I kept that logic/convention also for BTCPay plugin so you can configure a testnet and mainnet BTCPay server and switch between them easily (without pairing again on each switch). But yes, in an ideal world you have a test instance of the shop where testnet is connected and a live instance where mainnet is connected an you do not switch between both.

PR #22, opened 2018-08-18
Add Support Info

normandmickey:
Added support contact info


normandmickey:
I thought about it but wasn't sure if you would think it was too spammy.  I
will add it and create another PR.

PR #30, opened 2018-09-25
[docs] Add how to use existing bitcoin data_dir for the btcpay docker

BobleChinois:
I suggested in an [issue on btcpayserver-docker](https://github.com/btcpayserver/btcpayserver-docker/issues/34) that we find a way to use an already synced blockchain with BTCPay when possible.

The point is to allow people that already run a node to avoid a long and costly resync and leverage the work already done for their BTCPay full node.

I made a few modifications to `docker-compose.generated.yml`, and made BTCPay mount my existing `.bitcoin` directory instead of the docker volume `bitcoin_datadir`.

It doesn't seem to represent any risk, I was able to reuse the same directory with plain `bitcoind` thereafter, and then again with BTCPay without issue.

I propose this very easy "how to" as an addition to the FAQ, waiting maybe for a more durable solution one day.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -51,17 +51,19 @@ If you would like to run Bitcoin and Lightning Network nodes, the minimal requir
 
 ## I'm already running a full node and have a synched blockchain, can I make BTCPay use it so that it doesn't have to do a full sync again ?
 
-If you run BTCPay inside a docker compose, it is possible to bind docker's volume to a specific directory in your local host by modifying the file `docker-compose.generated.yml` inside of `btcpayserver-docker/Generated`. If the binded directory contains blockchain's data, bitcoind will recognize it and use it instead of doing the full sync again. 
+If you run BTCPay inside a docker compose, it is possible to bind docker's volume to a specific directory in your local host **without** modifying the file `docker-compose.generated.yml` inside of `btcpayserver-docker/Generated` using a _symlink_. If the directory your link is pointing at contains blockchain's data, bitcoind will recognize it and use it instead of doing the full sync again. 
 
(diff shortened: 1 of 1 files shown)


NicolasDorier:
this solution is not very good, using a softlink instead would not require renaming (https://stackoverflow.com/questions/9587445/how-to-create-a-link-to-a-directory)

BobleChinois:
Ok, just to be sure I got it right, what you mean is that I should create a symlink from docker's volume to my existing .bitcoin, so that I don't need to modify the .yml file, is that correct ?

BobleChinois:
I tried to do that, tracked the location of `generated_bitcoin_datadir` on my host with `docker volume inspect`, removed the automatically generated volume and created a symlink to my `.bitcoin` directory with `ln -s /home/sosthene/.bitcoin2 /var/lib/docker/volumes/generated_bitcoin_datadir`

Next time I started, it created a `_data` dir inside my `.bitcoin2`, and then all the directories (blocks, chainstate...) inside the `_data`. 

So I stopped it, rm all the content of the `_data` dir, and moved all of `.bitcoin2` inside `_data`.

I started again btcpay, and it seems to work this way.

I'm doing a few more test to be sure.

BobleChinois:
Yes, it should obviously work better. I'll do it with another machine to see if it still works. 

I have a question though, how is doing things this way better than modifying the .yml file ? Just to understand :smile:

BobleChinois:
I tested it with another machine and confirm that using symlink works for me. 

I propose another "how to" for the FAQ, reproducing the steps I have done myself.

NicolasDorier:
awesome

PR #31, opened 2018-10-01
Update Support.md

Kukks:


NicolasDorier:
We should advertise more about it everywhere

PR #33, opened 2018-10-03
[doc] update Support.md

BobleChinois:
I propose to modify the Support.md page as such as a quick fix for the issue with the Slack being used as a channel for support.

I found out that on Github's Support.md file, the NBitStack is mentionned only at the bottom, with the Slack, making it look like there are two equal alternatives for looking for support while they really aren't.

A lot of people are showing up on the Slack to get help, but it does not suit our need well : 
* Because of the message limit, it becomes harder and harder to find an issue that as already been answered a few days ago
* Searching for a specific info is a mess anyway !
NBitStack is much more suited for that, and already concentrate a great deal of issues and info that most users could browse themselves before getting to the Slack.

### DIFF
--- a/Support.md
+++ b/Support.md
@@ -8,7 +8,7 @@ Problem ? Don't worry, someone else has probably been through that before you.
 
 * First have a look at [NBitStack](https://nbitstack.com/c/btcpayserver) for a curated list of topics already answered by the community.
 
-* If your issue is not referenced there, or you want to feature request with BTCPay, please file a [github issue](https://github.com/btcpayserver/btcpayserver/issues).
+* If your issue is not referenced there, or you want to request a new feature, please open a [github issue](https://github.com/btcpayserver/btcpayserver/issues).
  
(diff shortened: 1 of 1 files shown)


Support.md
```diff
@@ -2,6 +2,18 @@
 
 This project is open source, and is not a company. Instead we rely on a network of contributors and users to provide support.
 
+## Free support
+
+Problem ? Don't worry, someone else has probably been through that before you. 
+
+* First have a look at [NBitStack](https://nbitstack.com/c/btcpayserver) for a curated list of topics already answered by the community.
+
+* If your issue is not referenced there, or you want to feature request with BTCPay, please file a [github issue](https://github.com/btcpayserver/btcpayserver/issues).
```
bitcoinshirt: If your issue is not referenced there, **or you want to request a new feature** please **open** a [github issue](https://github.com/btcpayserver/btcpayserver/issues).

bitcoinshirt:
ACK.

PR #61, opened 2018-10-22
Suggestions for "Use Case" page

BobleChinois:
I have some suggestions for the structure of the "Use Cases" page : 
* I feel it is more about "Value proposition" for users (principally merchants) and "opportunities" for other actors of the ecosystem to build upon BTCPay, so I reorganised the page in two sections and renamed it to "valueproposition.md" for short
* I added some more thoughts about why it is bad to reuse address, which is a strong argument in favor of BTCPay
* I made some minor editing for style

What do you think ?

### DIFF
--- a/UseCase.md
+++ b/UseCase.md
@@ -79,4 +79,4 @@ BTCPay community is amiable and open-minded. Developers can not only learn a lot
 
 ---
 
-These are some of the many ways in which you can use BTCPay. Unleash your creativity and feel free to build your own solutions that solve problems.
+These are some of the many ways in which you can use BTCPay. Unleash your creativity and feel free to build your own solutions to solve problems.
--- a/UseCase.md
+++ b/UseCase.md
@@ -16,7 +16,7 @@ By choosing BTCPay to process payments, you are:
 * Saving time (easy integration and installation)
 * Protecting yourself from interference in your business (self-sovereignty) 
 
-To enjoy most of these benefits, you don't even need to run a BTCPay server yourself, you could just create an account on someone else's server. It will be even easier to set up and entirely free, at the cost of relying on the server admin to keep it functional and up-to-date.  
+To enjoy most of these benefits, you don't even need to run a BTCPay server yourself, you could just create an account on someone else's server. It will be even easier to set up for free or a fee depending on the host's choice. Downside is that you will rely on the server admin to keep it functional and up-to-date.  
 
 With the growth of your business it will eventually become important to set-up your own server to be really independent.
 
--- a/UseCase.md
+++ b/UseCase.md
@@ -1,12 +1,10 @@
-# BTCPay Server: Value Proposition and Opportunities
+# BTCPay Server: Use Cases
 
 Build with the community in mind, BTCPay's ambition is to deliver the best Bitcoin payment service for merchants and their customers, while creating opportunities for everyone to contribute and take parts in its development and growth.
 
-## Value Proposition
+## Merchants
 
-### Merchants
-
-By choosing BTCPay to process their payments, you are:
+By choosing BTCPay to process payments, you are:
 
 * Saving money (no fees, no subscriptions)
 * Cutting off the middle-man (Payments go directly to your wallet)
@@ -14,13 +12,13 @@ By choosing BTCPay to process their payments, you are:
 * Saving time (easy integration and installation)
 * Protecting yourself from interference in your business (self-sovereignty) 
 
-#### Online Store
+### Online Store
 
 If you're a merchant running an e-commerce business, you can easily [deploy BTCPay](https://docs.btcpayserver.org/deployment) and connect it to your store via [integration plugins](https://docs.btcpayserver.org/integrations/) in just a few clicks. 
 
 BTCPay checkout is no different than any other payment gateway: your customer gets an invoice, that they pay by scanning a QR code or by copy-pasting the amount and the address in their own Bitcoin wallet. When their payment is confirmed, you will be notified via your e-commerce CMS; you can ship the item. Take a look at [our demo online store](https://store.demo.btcpayserver.org/).
 
-#### Physical Store
+### Physical Store
 
 For brick and mortar stores, BTCPay has a [web-based Point of Sale](https://mainnet.demo.btcpayserver.org/apps/87kj5yKay8mB4UUZcJhZH5TqDKMD3CznjwLjiu1oYZXe/pos) (POS) App which can be customized. Similarly to the online store, your customer is presented with an invoice that he can pay on the spot. You can create a watch-only wallet on your phone to be notified of the payments through the POS, without the need of any additional software. There are plans to develop the POS mobile app which would make the process even smoother.
 
@@ -28,7 +26,7 @@ For brick and mortar stores, BTCPay has a [web-based Point of Sale](https://main
 
 No matter the kind of business you run, BTCPay offers a very easy way to get started on the Lightning Network. You can use and experiment with this innovative second-layer solution build on top of Bitcoin by [following this guide](LightningNetwork.md). Both merchants and customers can use BTCPay to receive or make payments off-chain with instant confirmations and neglectable network fees.
 
-### Charities and Content Creators (Donations)
+## Charities and Content Creators (Donations)
 
 Charities, non-profits, content creators and other organizations that want to accept cryptocurrencies donations in a more private way than the traditional single-address can use the [Pay Button](https://docs.btcpayserver.org/btcpay-basics/gettingstarted#creating-the-pay-button) or the [POS app](https://github.com/btcpayserver/btcpayserver-doc/blob/master/GettingStarted.md#creating-the-point-of-sale-app).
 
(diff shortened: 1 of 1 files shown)


ValueProposition.md
```diff
@@ -38,7 +38,7 @@ Benefits of using BTCPay for accepting donations:
 * Cutting off the middle-man (Payments go directly to their wallet)
 * Enhancing privacy for them and their customers (no address re-use, no IP leaks to third parties)
 
-It is particularily important to mention that BTCPay prevents address reuse, as many people has been reusing address for donations in the past. Here is why SHOULD NOT reuse Bitcoin address:
+It is particularily important to mention that BTCPay prevents address reuse, as many people has been reusing address for donations in the past. Here is why you SHOULD NOT reuse Bitcoin address:
```
bitcoinshirt: This is good part!

UseCase.md
```diff
@@ -1,64 +1,71 @@
...
 
-### Physical Store
+### Lightning Network payments
 
-For brick and mortar stores, BTCPay has a [web-based Point of Sale](https://mainnet.demo.btcpayserver.org/apps/87kj5yKay8mB4UUZcJhZH5TqDKMD3CznjwLjiu1oYZXe/pos) (POS) App which can be customized. Similarly to the online store, your customer is presented with an invoice which needs to be paid. You can create a watch-only wallet on your phone to be notified of the payments through the POS, without the need of any additional software. There are plans on developing the POS mobile app which would make the process even smoother.
+No matter the kind of business you run, BTCPay offers a very easy way to get started on the Lightning Network. You can use and experiment with this innovative second-layer solution build on top of Bitcoin by [following this guide](LightningNetwork.md). Both merchants and customers can use BTCPay to receive or make payments off-chain with instant confirmations and neglectable network fees.
 
-## Charities and Content Creators (Donations)
+### Charities and Content Creators (Donations)
```
bitcoinshirt: This needs to be H2

bitcoinshirt:
While I agree with some changes, I don't like how "value proposition sounds".
The goal of the use-case page is clear :
1. Who can use btcpay
2. What they get out of it
No need to complicate things imho.

BobleChinois:
While I agree that "value proposition" might not be very satisfactory, I think it is not quite clear either to blend "real" users (merchants, charities and to some extent their customers/donators) and other actors like exchanges and hosting providers, that are more like possible contributors and partners in the development of BTCPay. 

Sorry it's not easy to review because I made the mistake to rename the file. I'll rollback and submit another version that are closest to your original idea.

bitcoinshirt:
> either to blend "real" users (merchants, charities and to some extent their customers/donators) and other actors like exchanges and hosting providers, that are more like possible contributors and partners

Valid point, however, all of them can **use** BTCPay in a way and BTCPay brings them a certain type of value. 

We can divide hosting providers and exchanges into another section, but I personally don't think that's of high importance. The use-case for each is clearly described. If you can think of how we can group "partners" as you call them from actual users more naturally, we can do it, but again, I think the current structure is concise enough and the page does what it should.

BobleChinois:
Ok I think I messed up with the git reset :man_facepalming: 

I'll close this PR and start a new afresh

UseCase.md
```diff
@@ -1,64 +1,82 @@
...
 
 ### Physical Store
 
-For brick and mortar stores, BTCPay has a [web-based Point of Sale](https://mainnet.demo.btcpayserver.org/apps/87kj5yKay8mB4UUZcJhZH5TqDKMD3CznjwLjiu1oYZXe/pos) (POS) App which can be customized. Similarly to the online store, your customer is presented with an invoice which needs to be paid. You can create a watch-only wallet on your phone to be notified of the payments through the POS, without the need of any additional software. There are plans on developing the POS mobile app which would make the process even smoother.
+For brick and mortar stores, BTCPay has a [web-based Point of Sale](https://mainnet.demo.btcpayserver.org/apps/87kj5yKay8mB4UUZcJhZH5TqDKMD3CznjwLjiu1oYZXe/pos) (POS) App which can be customized. Similarly to the online store, your customer is presented with an invoice that he can pay on the spot. You can create a watch-only wallet on your phone to be notified of the payments through the POS, without the need of any additional software. There are plans to develop the POS mobile app which would make the process even smoother.
+
+### Lightning Network payments
```
bitcoinshirt: This has to be a separate H2 headline since it's for both users who want to experiment and use LN and merchants. Do not mix these sections.
>

UseCase.md
```diff
@@ -1,64 +1,82 @@
-# BTCPay Server Use Case
+# BTCPay Server Use Cases
 
-While primarily build for merchants, BTCPay server is feature-rich software which has plenty of use-cases and solves problems for different users. In this guide, we will show you some of the use-cases of BTCPay. The usage of the software is not limited to these groups.
+Build with the community in mind, BTCPay's ambition is to deliver the best Bitcoin payment service for merchants and their customers, while creating opportunities for everyone to contribute and take parts in its development and growth.
```
bitcoinshirt: I don't like this part, it sounds like BTCPay is providing a service of some sort.

UseCase.md
```diff
@@ -1,64 +1,82 @@
...
+
```
bitcoinshirt: > why it is best for you!
Re-phrase this, sounds to cooky. Maybe : What are the benefits of using it.

UseCase.md
```diff
@@ -1,64 +1,82 @@
...
+If you're a merchant running an e-commerce business, you can easily [deploy BTCPay](https://docs.btcpayserver.org/deployment) and connect it to your store via [integration plugins](https://docs.btcpayserver.org/integrations/) in just a few clicks. 
+
+BTCPay checkout is no different than any other payment gateway: your customer gets an invoice, that they pay by scanning a QR code or by copy-pasting the amount and the address in their own Bitcoin wallet. When their payment is confirmed, you will be notified via your e-commerce CMS; you can ship the item. Take a look at [our demo online store](https://store.demo.btcpayserver.org/).
```
bitcoinshirt: > your customer gets an invoice**,** that they pay by scanning a QR code or by copy-pasting the amount and the address in their own Bitcoin wallet.

Re-phrase this, the comma should not be here. Make sentence shorter

UseCase.md
```diff
@@ -1,64 +1,82 @@
...
 
-Charities, non-profits, content creators and other organization that want to accept cryptocurrency donations in a more private way than the traditional single-address can use the [Pay Button](https://docs.btcpayserver.org/btcpay-basics/gettingstarted#creating-the-pay-button) or the [POS app](https://github.com/btcpayserver/btcpayserver-doc/blob/master/GettingStarted.md#creating-the-point-of-sale-app).
+Charities, non-profits, content creators and other organizations that want to accept cryptocurrencies donations in a more private way than the traditional single-address can use the [Pay Button](https://docs.btcpayserver.org/btcpay-basics/gettingstarted#creating-the-pay-button) or the [POS app](https://github.com/btcpayserver/btcpayserver-doc/blob/master/GettingStarted.md#creating-the-point-of-sale-app).
 
 Benefits of using BTCPay for accepting donations:
 
 * Saving money (no fees, no subscriptions)
-* Payments go directly to your wallet (No third-party)
-* Enhanced privacy for you and donors (no address re-use, no IP leaks to third parties)
+* Cutting off the middle-man (Payments go directly to their wallet)
```
bitcoinshirt: > to their wallet
Whose wallet?

BobleChinois: I hesitated between second and third person and switched many times, forgot to change this one. 

But there is another part where this is a real issue, it is the "local payment processor" part. I just realised I first talk from the perspective of the server admin that becomes payment processor for others, and then list benefits for other people to use a third party hosting. This is really not consistent...

I'm mentioning it because I think I could cut it off from there, but add a sentence in the first "merchants" part just to mention that BTCPay benefits are real even if you use third party hosting. 

I'll do a version like that so you can see what I mean

UseCase.md
BobleChinois: I see, I rolled back to something very close to what you originally wrote

UseCase.md
bitcoinshirt: I actually think you handled that quite well in the new version.

UseCase.md
```diff
@@ -1,64 +1,83 @@
-# BTCPay Server Use Case
+# BTCPay Server Use Cases
+  
+Build with the community in mind, BTCPay server is a feature-rich software with plenty of use-cases that can solve problems for different types of users.
```
bitcoinshirt: **Built** with a community in mind.

UseCase.md
```diff
@@ -1,64 +1,83 @@
...
-* Payments go directly to your wallet (No third-party)
-* Enhanced privacy for you and your customers
-* Easy integration and installation
-* Self-sovereignty 
+* Cutting off the middle-man (Payments go directly to your wallet)
+* Enhancing privacy for you and your customers (no address re-use, no IP leaks to third parties)
+* Saving time (easy integration and installation)
+* Protecting yourself from interference in your business (self-sovereignty) 
+
+To enjoy most of these benefits, you don't even need to run a BTCPay server yourself, you could just create an account on someone else's server. It will be even easier to set up and entirely free, at the cost of relying on the server admin to keep it functional and up-to-date.
```
bitcoinshirt: > It will be even easier to set **up and entirely free**, at the cost of relying on the server admin to keep it functional and up-to-date.

1. Not all hosts are free.
2. I wouldn't encourage the third-party hosting like this as people will get a false sense of security with server admin, but he can always close the server and leave them in the dark.

Suggestion remove this part 

> It will be even easier to set up and entirely free, at the cost of relying on the server admin to keep it functional and up-to-date.

bitcoinshirt:
Thanks for keeping up with my OCD @BobleChinois! I'm merging this now.

Issue #88, opened 2018-11-01
Re-Structure FAQ


bitcoinshirt:
Right now we have lots of good information scattered around and it would be a good idea to consolidate and categorize them.

My suggestion is to have three types of documents:

1. [FAQ.md ](https://github.com/btcpayserver/btcpayserver-doc/blob/master/FAQ.md)- General BTCPay questions which are more about navigation, options, features.
2. Troubleshooting.md #48 - Document that will describe how users can troubleshoot a certain problem and get help. This is more general information document rather than coding instructions.
3. Common Issues - This will be a huge document with all the technical issues categorized. The goal is to provide code so that people can use. I propose we take a look at FAQ and the Wiki to create categories and structure the questions. For instance, - Synchronisation - all questions about the sync (why it's needed, how long it takes, what to do if it's stuck.

The 3. document will potentially take time and we'll need to think of the "categories" for the issues. The easiest is to just take a look at NBitstack Wiki and try to group similar issues.

Thoughts?

bitcoinshirt:
## Wallet
 
- [Can I use Bitcoin Core wallet with BTCPay Server?](https://nbitstack.com/t/can-i-use-a-bitcoin-core-wallet-with-btcpayserver/33)
- [What gap limit is optimal?](https://nbitstack.com/t/how-to-make-sure-that-the-gap-limit-in-btcpay-is-picked-up-by-my-wallet/59)

## Apps

- [How can I use BTCPay in a physical store?](https://nbitstack.com/t/how-can-i-change-the-currency-from-usd-to-some-other-fiat-currency-and-use-btcpay-as-a-physical-pos/38)
- [Is there a Point of Sale feature in BTCPay?](https://nbitstack.com/t/is-there-a-pos-feature-for-btcpayserver/47/3)
- [How to get sucessful payment notification if I'm not using e-commerce software](https://nbitstack.com/t/how-can-i-be-notified-about-btcpay-payment-if-im-not-using-e-commerce-software/63/2)

## Lightning Network

### LND
- [How many Zap Wallets can I connect to BTCPay at simultaneously?](https://nbitstack.com/t/how-many-zap-wallets-can-be-connected-to-a-btcpay-server-ln-simultaneously/45/4)
[...]

PR #90, opened 2018-11-03
Creating /FAQ and FAQ-Synchronization.md

bitcoinshirt:
This PR is an attempt to solve a portion #88 

The idea is to have a FAQ folder where we can categorize issues. For each major category, we create a document with a Table of Content (TOC) and list all of the common questions.

This PR specifically creates a FAQ folder and adds a new document FAQ-Synchronization.md. 
FAQ-Synchronization.md contains frequently asked questions about BTCPay sync.

My idea is to consolidate all the issues by categories [like in this example](https://github.com/btcpayserver/btcpayserver-doc/issues/88#issuecomment-435023619) into a single FAQ folder. FAQ should also have its own readme, linking to everything for easier navigation.

I intend to add FAQ-General which will pretty much be the FAQ.md we already have minus the non-general type of questions. More technical issues should be consolidated in separate documents like FAQ-Synchronization.md.

For end-user, this does not change much in terms of navigation. With our new GitBook upgrade, users will easily be able to find anything with a search function on docs page

This PR is a bit tricky and changes the structure, so your feedback is welcome.


bitcoinshirt:
Thoughts on structuring these sort of technical issues @britttttk @BobleChinois?

britttttk:
Making FAQ into a folder is a good idea. Would be nice to see it at the top of the docs directory. 

For the FAQ-Sync.md file, I think if there is enough information (ex: questions/answers) to create a page then sure. If it's only a link to one post on NBitStack then maybe it can just be a link in your TOC page. 

I think that if we create folders (perhaps even 3: FAQ, Troubleshooting, CommonIssues) then we should keep a TOC for each one in README or whatever. 

Maybe it's best to make the FAQ folder & README/TOC first to break up the PR.

bitcoinshirt:
Good idea @britttttk , let me try that. I agree that if we plan to branch things out, we have to put proper navigation and linking so that users can find the question easier.

bitcoinshirt:
So, should I rename the FAQ-sync.md into **Issues-Sync.md**, rename the **folder** as common-issues and keep the FAQ as a page with general questions only, while we keep technically advanced issues into the common-issues folder?

Or, do we merge everything into FAQ, and have FAQ page as FAQ-General and technical issues like FAQ-Sync.md

britttttk:
@bitcoinshirt I think you are right, and as you mentioned in #88 it would probably be best to split FAQ (general Q's) and CommonIssues (Tech Q's). I'm also thinking folders for these. I'm more used to using NBitStack so I'm not sure exactly about how it all works with GitBooks and layout.

Maybe something like this:

Issues (F)
    - TOC/README.md (Includes some kind of issue List with categorized links)
    - Troubleshooting.md (general info whatever)
    - Issues-Sync.md (Ex: Page due to numerous Q/A for one topic or very frequent issue)

-FAQ (F) 
    - TOC/README.md (Includes some kind of List with categorized links)
    - FAQ-LogoImages.md (Ex: Page with BTCPay logos)

bitcoinshirt:
I've been thinking about it for the whole day. I am now 100% leaning towards having a huge FAQ page with a document for each category, completely merging FAQ and Common issues.

After taking a closer look at issues users face, most of them can be classified as "technical"

So far I have : 
- /FAQ folder
- Inside FAQ [I have FAQ readme](https://github.com/btcpayserver/btcpayserver
-doc/blob/b40efcea5db4e8f630a12e6cf9025c9f001c302d/FAQ/readme.md) which is a huge list of all issues that will help anyone quickly find an answer even with simple command + F on a browser. 
- FAQ General- contains general non-technical issues
- FAQ Sync - technical and less technical issues about sync
- FAQ Wallet - technical and less technical issues about sync

I think I'll start this PR with consolidating as many issues into categories. People are more likely to read Snyc.md than about sync in FAQ and sync in "common issues". 

Take a look at PR and let me know what you think about this new structure.

britttttk:
> After taking a closer look at issues users face, most of them can be classified as "technical"

^Ahh ok yes I didn't think about this, good idea to keep it combined.

I'm 100% on the folder and think your layout looks great.

NicolasDorier:
this seems awesome to me. We start to get too much FAQ, it makes sense to break them down like this.

bitcoinshirt:
Okay, merging this one and from here. will submit tiny PR for each category.

Issue #57, opened 2018-10-20
[New Doc] Cryptocurrency Payment Processors - Comparison


bitcoinshirt:
I remember @NicolasDorier had an idea of having a chart comparing different payment processors to BTCPay.

I think it can be really helpful to have such a document for the upcoming users. The biggest problem is figuring out what to compare and what features to include.

If anyone has an idea or wants to tackle this, feel free to comment.

geraldz:
1.  Fees
2.  Ability to transfer BTC to USD and transfer to bank account (and related fees)
3.  Personal use only or available for merchants?

bitcoinshirt:
Good idea @geraldz  here are a few more ideas:

1. Fee Structure
2. Available Integrations
3. Supported Coins
4. Requires Personal Information
5. Middleman involvement
6. Fiat Conversion
7. Lightning Network Support
8. Non-custodial wallet

britttttk:
I found [this website](https://www.g2crowd.com/categories/cryptocurrency-payment-apps) it has a list of them... and we are on it. Should i give us a review :)

r4victor:
I'm working on such a comparison. One may want to use it as a reference. You can find it [here](https://howtoacceptcrypto.com/chart/).

bitcoinshirt:
Sweet @r4victor this is so cool. Excellent work. 
If you wish to have this published here in the docs or on btcpay site, feel free to create a new doc.
Excellent.

r4victor:
Thanks, @bitcoinshirt! The work is still in progress, so I don't think this chart should be embedded in the docs. It can though be a link in, for example, "Why choose it over other payment processors?​" section, but I'm not sure whether it's appropriate to reference an external resource or not.

britttttk:
@r4victor thanks for the comparison info nice work!! @bitcoinshirt I agree, I think we definitely have have enough info to make the doc. I can work on it.

PR #130, opened 2018-12-06
Payment processor comparison

britttttk:
Draft of #57:

- Compare BTCPay as a payment processor
- Explain processor trade-offs
- Link BIP70 converter

Reasons:
I made a chart to compare processors. Some of the websites make some claims I thought were not true (mostly for KYC and middleman). So I updated the table, but then it goes against what they advertise so I also don't want to get sued for defamation.... so instead of putting their names in a table, I just identify them by what they do.

### DIFF
--- a/ComparePaymentProcessors.md
+++ b/ComparePaymentProcessors.md
@@ -1,0 +1,95 @@
+# Payment Processor Comparison
+
+Most new merchants will likely only consider the price of the service. Since BTCPay is free, that may have led you here and if so, welcome. 
+
+We want users to understand not only BTCPay, but also how payments can be processed on behalf of a merchant. In order to help users understand trade-offs when using various cryptocurrency payment processing methods. We don't know which payment processors provide which services. That will need to be part of your own research. The following list is a good place to start:
+
+* [Features](#Features)
+* [Security](#Security)
+* [Privacy](#Privacy)
+* [Decentralized](#Decentralized)
+* [Fiat](#Fiat)
+
+----
+
+## Features
+
+Every payment processor has features, here are some BTCPay features:
+
+* **Bitcoin** - Accepting Bitcoin is the first step.
+* **Altcoins** - Accept cryptocurrency alternatives to Bitcoin.
+* **Lightning** - Rapid Bitcoin microtransactions using the Lightning Network.
+* **Integrations** - Connect to your website using WooCommerce, Drupal, Magneto, Prestashop and custom integrations.
+* **Point Of Sale** - POS Interfaces for physical stores.
+* **Unlimited Stores** - Merchants can process payments for their own stores, or for others.
+* **Merchant Support** - 24/7 open source community support.
+* **Translations** - Customers can pay in 20 different languages. 
+
+----
+
+## Security
+
+First rule of Bitcoin is always keep your private keys *private*. Using a secure wallet is recommended for new merchants as the only provider (creator) of private keys. If there is a chance that someone else (such as a website) knows, stores, or provides your private keys to you, it's generally accepted that they are not actually private. 
+
+Secondly, there is another area of security to consider on the applications layer where you have two main options:
+
+* **Option 1**: Most payment processors (including BTCPay) use the [BIP 21][2] standard.
+* **Option 2**: Others use variations of the [BIP 70][3] standard.
+
+     * **Note**: [BIP 70 has recently been depreciated in Bitcoin Core][1].
+     * Many wallets do not allow payments to BIP 70 invoice urls.
+     * Need help converting obsolete BIP 70 invoice urls to a BIP 21 invoice url? No problem, we can help with that too. <br>Here's our [BIP 70 invoice url converter][4].
+
+----
+
+## Privacy
+
+BTCPay will never ask a merchant for any personal identification other than your email address. 
+
+Typically when converting to or from fiat on behalf of a merchant, payment processors are required to collect personal information for Know Your Customer (KYC) and Anti-money laundering (AML) banking requirements. This may include personal information such as passport ID, phone number, address, bank account, etc. 
+
+Fortunately, the Bitcoin Network does not use or collect these types of personal information, and therefore neither does BTCPay. 
+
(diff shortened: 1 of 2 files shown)


ComparePaymentProcessors.md
```diff
@@ -0,0 +1,95 @@
...
+
+     * **Note**: [BIP 70 has recently been depreciated in Bitcoin Core][1].
+     * Many wallets do not allow payments to BIP 70 invoice urls.
+     * Need help converting obsolete BIP 70 invoice urls to a BIP 21 invoice url? No problem, we can help with that too. <br>Here's our [BIP 70 invoice url converter][4].
+
+----
+
+## Privacy
+
+BTCPay will never ask a merchant for any personal identification other than your email address.
```
bitcoinshirt: This can be confusing. What are you referring to here exactly?
BTCPay never asks for any emails, this sounds like BTCPay is a third party which collects emails.

ComparePaymentProcessors.md
```diff
@@ -0,0 +1,95 @@
...
+* [Privacy](#Privacy)
+* [Decentralized](#Decentralized)
+* [Fiat](#Fiat)
+
+----
+
+## Features
+
+Every payment processor has features, here are some BTCPay features:
+
```
bitcoinshirt: You forgot the biggest feature which will interest any merchant - no fees. I know you mentioned in the first sentence of the document, but we should clarify why is it free, what sort of fees are not charged.

ComparePaymentProcessors.md
```diff
@@ -0,0 +1,95 @@
...
+That's why BTCPay was created. To help merchants remove third party dependencies and simply use the Bitcoin Network freely and securely. Merchants have a copy of the BTCPay software which runs on their own server or VPS of their choice and validates their own payments using their own node. It's a self-hosted Peer-to-Peer payment processor. The trade-off in this case is some technical understanding is required for initial setup.
+
+As the BTCPay community continues to grow, more deployment methods, use cases and tutorials are continually being added to make it easier for non-technical users. BTCPay is completely open source. Anyone can join the community to suggest or create improvements, features, guides, etc. Feedback is always welcome. 
+
+----
+
+## Fiat
+
+Currently, BTCPay is a processor **without fiat conversion** capability. As a merchant, this may be a difficult if business costs require fiat. Not providing fiat conversion allows BTCPay merchants to avoid KYC and AML identification verification. This also allows BTCPay to be free and availible for anyone to use.
```
bitcoinshirt: However, mention that it's something on the roadmap and that since merchants are always the owners of their public keys, hence the coins, they can always freely convert their coins manually, but for now there's no instant-fiat conversion.

ComparePaymentProcessors.md
```diff
@@ -0,0 +1,95 @@
+# Payment Processor Comparison
+
```
bitcoinshirt: I think this can be added as an introduction https://github.com/btcpayserver/btcpayserver-doc#how-is-it-different

ComparePaymentProcessors.md
bitcoinshirt: Rather than focusing on email, I would focus:
- BTCPay is private because there's no middleman
- Any information is shared only between the seller and a customer
- If merchant runs a self-hosted BTCPay, that means they are running a full Bitcoin node, which is the ultimate privacy since no middleman can listen to their transactions. This is neat resource https://en.bitcoin.it/wiki/Why_Your_Business_Should_Use_a_Full_Node_to_Accept_Bitcoin
- No address re-use

ComparePaymentProcessors.md
```diff
@@ -0,0 +1,95 @@
...
+## Features
+
+Every payment processor has features, here are some BTCPay features:
+
+* **Bitcoin** - Accepting Bitcoin is the first step.
+* **Altcoins** - Accept cryptocurrency alternatives to Bitcoin.
+* **Lightning** - Rapid Bitcoin microtransactions using the Lightning Network.
+* **Integrations** - Connect to your website using WooCommerce, Drupal, Magneto, Prestashop and custom integrations.
+* **Point Of Sale** - POS Interfaces for physical stores.
```
bitcoinshirt: Include payment button?

ComparePaymentProcessors.md
```diff
@@ -0,0 +1,95 @@
...
+* **Unlimited Stores** - Merchants can process payments for their own stores, or for others.
+* **Merchant Support** - 24/7 open source community support.
```
bitcoinshirt: I would remove this, we can't guarantee any sort of support, especially not 24/7, I don't want people to take that for granted.

ComparePaymentProcessors.md
britttttk: Great ideas, I will add them thanks. The email i was referring to is when merchants log into BTCPay but I agree it is confusing. I will remove that.

ComparePaymentProcessors.md
britttttk: @bitcoinshirt it's just like how open source works. I'll remove it though. Otherwise everyone will be asking for you 24/7 :)

ComparePaymentProcessors.md
britttttk: @bitcoinshirt When you say payment button, are you talking about how merchants will have a "pay with Bitcoin" button shown on their website/POS? Meaning it adds another payment method, right?

bitcoinshirt: By Payment button, I mean the payment button from the store settings which can be embedded into HTML. It's like an additional app, like PoS. https://store.demo.btcpayserver.org/?page_id=44
So it's just another app, hence the feature, mostly used for donations, but merchants (like CasaHodl for example, use it their e-commerce solutions as well).

ComparePaymentProcessors.md
Eskyee: well spotted, I shall also remove from btcpay jungle as well, like you said 
"I don't want people to take that for granted"  
Yes, Our Support at BTCPay Community is the best for sure.

PR #154, opened 2019-01-07
Simplify raspberry steps

NicolasDorier:
Ping @normandmickey can you check if this work? I prefer removing manual steps for the user.


NicolasDorier:
I fixed the instruction to use sudo properly

normandmickey:
Will test it out this evening.

normandmickey:
Need to add "-a" to tee command to append the line.  Without it the whole file will be replaced with one line. 

echo "UUID=$UUID /mnt/usb ext4 defaults,nofail 0" | sudo tee -a /etc/fstab

NicolasDorier:
Thanks just fixed, the line is correct now?

normandmickey:
Yes, thanks!

Issue #198, opened 2019-03-18
Document TOR support


pavlenex:
Not sure if this should be under features or just FAQ. Seems better suited for FAQ imo.
Some information available here https://medium.com/@BtcpayServer/about-tor-and-btcpay-server-2ec1e4bd5e51 and here https://twitter.com/BtcpayServer/status/1107591715685167104

Whoever is familiar with TOR and uses it with btcpay, feel free to document it.

BobleChinois:
I just tried to activate tor, it was definitely very easy and it works well but I think it definitely needs some clarification about what it does and what it doesn't in the FAQ, because even after reading Nicolas' article on Medium I was a bit confused about it and thought it would force all bitcoind and lnd trafic through Tor. 
If nobody's already working on it I can make a pull request soon.

pavlenex:
@BobleChinois Sure, go ahead, assigned this one to you. Ty.

PR #218, opened 2019-04-23
General deployment faq/tor

BobleChinois:
This is a proposal to solve issue #198.
In my view, there are 2 goals:
* explain how to activate Tor
* make it clear that it is not intended as a privacy and/or security scheme, but more like an improvement of the setup that would allow to run BTCPay on environment where it would have been hard otherwise because of some network requirements (as Nicolas has explained in a Medium article about this topic)
I think that it's enough to answer those 2 questions separately in the FAQ. I think the "General Deployment" FAQ was the more appropriate.
Please let me know if I got something wrong, or if the message could be made even more clear.

### DIFF
--- a/FAQ/FAQ-Deployment.md
+++ b/FAQ/FAQ-Deployment.md
@@ -113,9 +113,11 @@ Tor just solves all these issues in one shot, all you have to do is plug your de
 
 But if you're looking for perfect privacy and security, **activating Tor with your BTCPay just won't do it.** 
 
-Tor is a really tricky software to use for developpers, as the slightest mistake can tear down the anonymity it provides. As BTCPay is evolving into a rather complex service and adding more and more plugins, even if we tried to route all this trafic thourgh Tor, we can't guarantee that there will never be leaks of data in clear.
+Tor is a really tricky software to use for developpers, as the slightest mistake can tear down the anonymity it provides. As BTCPay is evolving into a rather complex service and adding more and more plugins, even if we tried to route all this trafic through Tor, we couldn't guarantee that there will never be leaks of data in clear.
 
 We think that the illusion of security is more dangerous that no security, or at least security we know is imperfect. So be aware that activating Tor doesn't prevent others to connect to your instance website, your bitcoin or lightning node in clear, **it doesn't make you anonymous at all.**
+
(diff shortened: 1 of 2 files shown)


pavlenex:
I'm not an expert for Tor, I think it's okay, but I'd need @NicolasDorier to have the final word on it. The place for it is appropriate imo. Just make sure to add links to FAQ in [FAQ readme,](https://github.com/btcpayserver/btcpayserver-doc/blob/master/FAQ/readme.md) as that's a Table of Content for all questions.

FAQ/FAQ-Deployment.md
```diff
@@ -79,6 +81,44 @@ sudo ln -s /mnt/usb/docker /var/lib/docker
...
+
+That's really easy: just log in your instance with SSH, and enter the `root/btcpayserver-docker` directory as root. There, type the two following command lines:
+```
+root/btcpayserver-docker $ BTCPAYGEN_ADDITIONAL_FRAGMENTS="$BTCPAYGEN_ADDITIONAL_FRAGMENTS;opt-add-tor"
+root/btcpayserver-docker $ . btcpay-setup.sh -i
+```
+
+Then wait a few minutes for the server to restart, and you're done!
+
+If you used to have other options passed with BTCPAYGEN_ADDITIONAL_FRAGMENTS, remember to add those as well. For example, if you pruned the blockchain, you need to pass this line instead of the one above:
```
NicolasDorier: no you don't need `BTCPAYGEN_ADDITIONAL_FRAGMENTS="$BTCPAYGEN_ADDITIONAL_FRAGMENTS;opt-add-tor"` already keep what you had before.

BobleChinois: Ah? First time I did that, and bitcoind got stuck, looking at the output it was asking to be restarted with a -reindex, so I thought it was maybe because I didn't pass the prune argument again, so I did just that and it was back on track. But maybe I was having another issue then...
Would it help if I turn off Tor and try to activate it again just to check if I can reproduce the error?

NicolasDorier: Yeah, give it a try, maybe you did typo.

NicolasDorier:
Awesome, looks good to me @pavlenex

Issue #214, opened 2019-04-16
Add Documention to Remove certain add-ons


Freefromchainz:
In the Process of trying to trim the fat on my server. Certain add-ons like libre patron that I'am not currently using i do not see documentation on how to remove them, as they maybe costing extra burst cpu usage on lunanode.

pavlenex:
ping @JeffVandrewJr you may want to document how to comment out in LibrePatron docs as well :) I'll do it in FAQ.

JeffVandrewJr:
I'm assuming we're talking about deactivating a previously activated instance, since if it's never active in the first place, this shouldn't be an issue. 

Rather than commenting out, I think it's cleaner to do the following:

`echo $BTCPAYGEN_ADDITIONAL_FRAGMENTS`

Then copy the entire string that pops up. 

Then do:

```
export BTCPAYGEN_ADDITIONAL_FRAGMENTS=<whole string other than opt-add-librepatron

. btcpaysetup.sh -i
```

Then you're permanently deactivated. 

If you want to take it a step further and destroy the data:

```
rm -r /var/lib/docker/volumes/librepatron
```

Freefromchainz:
Thank you jeff!

BobleChinois:
I just ran into this issue (needed to deactivate Tor to make a test), and figured how to do it thanks to this issue. Should I add it to the FAQ?

pavlenex:
@BobleChinois sure, feel free to add instructions in FAQ on how to deactivate/remove the plugin(s).

pavlenex:
[Added.](https://github.com/btcpayserver/btcpayserver-doc/pull/220)

PR #220, opened 2019-04-24
General deployment faq/del opt

BobleChinois:
A simple proposal for the FAQ based on @JeffVandrewJr answer to issue #214

### DIFF
--- a/FAQ/FAQ-Deployment.md
+++ b/FAQ/FAQ-Deployment.md
@@ -17,7 +17,7 @@ Here are common questions about installation, regardless of the deployment metho
 * [With the docker deployment, how to use a different volume for the data?](FAQ-Deployment.md#with-the-docker-deployment-how-to-use-a-different-volume-for-the-data)
 * [How do I activate Tor on my BTCPay Server?](FAQ-Deployment.md#how-do-i-activate-tor-on-my-btcpay-server)
 * [Why activate Tor? Does it mean that nobody knows who I am?](FAQ-Deployment.md#why-activate-tor-does-it-mean-that-nobody-knows-who-i-am)
-* [How can I modify/deactivate environment variables?](FAQ-Deployment.md#How-can-i-modify-deactivate-environment-variables)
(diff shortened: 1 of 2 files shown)


FAQ/FAQ-Deployment.md
```diff
@@ -17,6 +17,7 @@ Here are common questions about installation, regardless of the deployment metho
 * [With the docker deployment, how to use a different volume for the data?](FAQ-Deployment.md#with-the-docker-deployment-how-to-use-a-different-volume-for-the-data)
 * [How do I activate Tor on my BTCPay Server?](FAQ-Deployment.md#how-do-i-activate-tor-on-my-btcpay-server)
 * [Why activate Tor? Does it mean that nobody knows who I am?](FAQ-Deployment.md#why-activate-tor-does-it-mean-that-nobody-knows-who-i-am)
+* [How can I deactivate Tor (or any other option)?](FAQ-Deployment.md#how-can-i-deactivate-tor-or-any-other-option)
```
pavlenex: `or any other option` is unclear, nobody will search for how to deactivate any other option. I think you should re-write this to be plugin neutral and provide Tor as an example.

pavlenex: I think:
How to deactivate environment variable in BTCPay?

BobleChinois: Mmmhhh not sure if "environment variable" will be very clear for the average user. Do we have a complete list of all the possible options somewhere?

BobleChinois: otoh people that messing with the command line like this probably know about it

britttttk: yes here is the complete list: https://github.com/btcpayserver/btcpayserver-docker#environment-variables

BobleChinois: Ok cool, seems like we could add `opt-add-tor` to this list. Thanks! I'll get back to this later

britttttk: No, opt-add-tor is below under the BTCPAYGEN_ADDITIONAL_FRAGMENTS list. Might be a good idea to link the "Warning" to your "Why activate Tor?" though :)

Issue #227, opened 2019-05-08
Update walkthrough guide with new header image


pavlenex:
[Walkthrough guide](https://docs.btcpayserver.org/btcpay-basics/walkthrough) should be updated with new header image (remove old one) and new options/settings/features should be documented.

britttttk:
- [ ] Update header image
- [ ] Server Settings
- [ ] Stores
- [ ] Apps
- [ ] Wallets
- [ ] Invoices
- [ ] Payment Requests
- [ ] My Settings

PR #232, opened 2019-05-09
Update Walkthrough

britttttk:
#227


Walkthrough.md
```diff
@@ -71,4 +72,4 @@ For more information, check [Payment Requests](PaymentRequests.md)
 
 ## My Settings
 
-My Settings are user settings. Modify your e-mail address, password or set up two-factor authentication.
+My Settings are user settings. Modify your e-mail address, password or set up two-factor authentication. The U2F feature allows users to further enhance the security of their accounts by physically confirming wallet authentication via a U2F compatible device.
```
pavlenex: > by physically confirming wallet authentication via a U2F compatible device.

Hm, this part is a bit unclear, as U2F has nothing to do with a wallet.

Maybe:

> The U2F feature allows users to further enhance the security of their accounts by physically confirming authentication with a U2F compatible device.

I like this simple definition ([source](https://hackernoon.com/what-the-heck-is-u2f-35cb68082dbe))

> U2F is a physical key that you put into a USB port on your computer. You put this in after inputting your password. The U2F device uses encryption, as it contains a private key that is matched up to your public key in order to unlock your accounts like Gmail and Facebook. Without the physical key, no one can access your account. So, hackers, and even key loggers will not be able to steal your passwords because the U2F encrypts the data.

So maybe good to play with it and explain it better

PR #255, opened 2019-06-19
Change to the better php client

Kukks:


Kukks:
I feel ignored @pavlenex

Issue #256, opened 2019-06-23
Explain how to connect to remote LN node with Third Party host


britttttk:
Probably would be best in ThirdPartyHosting.md or an FAQ question.

pavlenex:
In 128 we've added the ability to share LN admin node with non-admins, so probably should be added as well somewhere in docs.

https://blog.btcpayserver.org/1-0-3-128-released/

PR #257, opened 2019-06-23
Explain remote LN node options in ThirdPartyHosts.md

reablaz:
According to [#256](https://github.com/btcpayserver/btcpayserver-doc/issues/256)

I've added a small notice with links (where is available full setup documentation, which is not related to btcpayserver itself) about how to connect remote c-lightning node.

### DIFF
--- a/ThirdPartyHosting.md
+++ b/ThirdPartyHosting.md
@@ -33,6 +33,9 @@ Yes. Here are some restrictions.
 
 The limitations happen for technical reasons, mostly because these features require a user to run a full node to use them.
 
+*Tip: If you are owner of [c-lightning](https://github.com/ElementsProject/lightning) Lightning Node you are able to create Lightning invoices with third-party host using [lightning charge drop-in](https://github.com/ElementsProject/lightning-charge). [Add remote lightning node](LightningNetwork.md) specifying type(charge), server and api-token specific for your own server.* 
+
+
 ## Pros
 * Easy and quicker setup
 * Cheaper and in most cases free (depending if the host is premium or free)


ThirdPartyHosting.md
```diff
@@ -33,6 +33,9 @@ Yes. Here are some restrictions.
 
 The limitations happen for technical reasons, mostly because these features require a user to run a full node to use them.
 
+*Tip: If you are owner of [c-lightning](https://github.com/ElementsProject/lightning) Lightning Node you are able to create Lightning invoices with third-party host using [lightning charge drop-in](https://github.com/ElementsProject/lightning-charge). [Add remote lightning node](LightningNetwork.md) specifying type(charge), server and api-token specific for your own server.*
```
pavlenex: Why only c-lightning? Maybe just tell people that if they have externally hosted lightning node, they can connect remotely, but that it's a technical step which does not make much sense nor is documented (it does not make sense because if you're already hosting a node, it's better to host btcpay itself which comes bundled up with everything. 
Not a fan of current explanation.

reablaz: I should mention also other implementations. I mentioned c-lightning because thats only one i've tried myself and it worked for me well.
Well, there are number of reasons why you may want use remote btcpayserver: 

- you are LN node operator and want to try btcpayserver for testing purposes
- (my case) you have very limited resources on bitcoin&LN node host or you are not able to run btcpayserver on same host for other technical reson
- you cannot or won't keep your own btcpayserver up to date for any reason
- you would like to use shared LN node
- you want to use 3rd party btcpayserver as a backup

i think there may be even more reasons. i personally find remote LN node feature very useful for my self and in general. So, how do you see it would be better to mention that such option is actually available with 3rd party server or shouldn't we?

pavlenex: People using third-party hosts are not technically knowledgable and are looking for a quick, cheap and easy way to accept BTC. I'm not sure that mentioning it more than one sentence makes sense. 

If you wish to further technically expand it, I think it's better if that explanation is placed in FAQ/ FAQ-Lightning Network, so I'd personally keep it short and simple in this one.

britttttk: I think it's quite possible that some technical people are using 3P hosts for development purposes. I like the explanation as it is now however I do agree with @pavlenex concerns because majority (by far) of people visiting this page are very new users who do not really understand much but are curious about it. So perhaps we can compromise with something like this:

_Note: It is technically possible to connect to remote nodes with third-party hosting services. However these methods require developer skills and documentation is currently limited, [see this example]. For that reason we recommend either deploying your own server to have full capability or utilizing a third-party host with some limitations._ 

FAQ-SeeThisExample:
How to connect third party host to lightining node?
The general process is that you need to ....  [ insert some > instructions > like this > btcpay page navigation ]. As an example, you can connect to a remote c-lightning node by using .... [insert specific steps or links to documentation / some text charge drop-in. Add remote lightning node (using general steps above) specify type as charge, something something server and api-token specific for your own server. ]

pavlenex:
As a part of pull-request cleanup I am closing this PR since it seems there's a better way to approach this which needs to be way more detailed.

PR #268, opened 2019-07-15
Added FAQ item "Can I use an existing Nginx server as a reverse proxy with SSL termination?"

woutersamaey:
Added FAQ item "Can I use an existing Nginx server as a reverse proxy with SSL termination?"


NicolasDorier:
Cool! Can you add a notice though? If the install has more than one domain (for example `WOOCOMMERCE_HOST` or `BTCPAY_ADDITIONAL_HOSTS`) you are responsible to also modify your reverse proxy.

woutersamaey:
Done

pavlenex:
@NicolasDorier review and merge if it's okay please.

PR #274, opened 2019-07-24
Update Raspberry Pi Install Instructions

normandmickey:
Updated Raspberry Pi installation instructions to accommodate Docker on Raspbian Buster.


RaspberryPiDeployment.md
```diff
@@ -142,23 +142,31 @@ sudo mkdir /mnt/usb/docker
...
 ```bash
 sudo /etc/init.d/dphys-swapfile stop
 sudo /etc/init.d/dphys-swapfile start
 ```
 
-**Step 14** - Install BTCPayServer.  
+**Step 14** - Install Docker. 
+Docker-CE is not available yet for Raspbian Buster.  Until Docker is available in the official repo it has to be installed manually. 
+```bash
+curl -fsSL https://get.docker.com|sed -e 's/buster/stretch/' > get-docker.sh
```
jmaurice: It's patched in the setup script so don't need to update the installation docs with the patch https://github.com/btcpayserver/btcpayserver-docker/pull/176

RaspberryPiDeployment.md
```diff
@@ -13,7 +13,7 @@ The process is basically the following:
 
 BTCPayServer can be successfully installed on the following hardware.   
  
-1. Raspberry Pi 3 Model B+
+1. Raspberry Pi 3 Model B+, Raspiberry Pi 4 Model B
```
wiz: I think what we're going to do is make a separate document for RPI4 because the instructions have some significant differences

normandmickey: Ok. It may not even be worth mentioning the Raspberry Pi 3 anymore.  BTCPayServer can be installed on it but memory is an issue.  BTCPayServer works great 2GB but obviously 4GB is better.;

wiz:
I want to make some tweaks but I'll submit a separate PR so let's merge this one for now. Thanks @normandmickey for your contribution!

PR #275, opened 2019-07-24
howtoacceptcrypto.com do not exist anymore

HugoDoyon:
howtoacceptcrypto.com website do not exist anymore (Dead for more than a week at least)


HugoDoyon:
indeed, will awesome list are a great thing

pavlenex:
Thanks for the PR @HugoDoyon, good catch.
@britttttk I like that list you linked, had the exactly the same idea, let's go with it instead.

PR #278, opened 2019-07-25
Add another way to solve reindex-chainstate

pavlenex:
This is apparently another way to tackle reindex-chainstate problem with corrupted nodes.
I have not tested it, but a community member cr0wn_gh0ul reported it worked for their use-case.

It's untested, so requires a review and approval of someone more familiar with the problem.


NicolasDorier:
I would be surprised if it works, if the main process die this would close this bitcoind as well.

pavlenex:
So NACK? I really have no way of testing it, so whatever you say on this one.

NicolasDorier:
@pavlenex most of the time the error `reindex-chainstate` come because people are using FastSync but forget to use `opt-save-storage`.

pavlenex:
@NicolasDorier I saw that reply on Mattermost, will add another PR and better explain that.

Issue #213, opened 2019-04-14
xpub generation video and doc update


pavlenex:
People who don't have Nano S or Trezor are having a hard time finding a way to have p2sh compatible xpubkey.

Unfortunately, it's not possible to generate a p2sh seed in Electrum, you can only import it from the external wallet, so we need to create a guide for those people, as I'm seeing an increasing amount of confused newbies.

Any wallet suggestions or workarounds @NicolasDorier?

pavlenex:
Two wallets me and @rockstardev found would be Samourai for Android and Coinomi for iOS. However, neither supports custom gap limit which will be a deal breaker and will cause "where's my money" questions...

So how do we make this easier for people who don't have or can't afford a hardware wallet and for those who want to track payments externally?

I'm really hesitant to open a feature request for watch-only BTCPay Wallet app...

This issue is more for the main repo but maybe we can brainstorm a bit here.

JeffVandrewJr:
This can be run fully client-side without an internet connection using a local copy: https://github.com/iancoleman/bip39

The BIP49 in the generator tab generates P2SH wrapped Segwit addresses.

pavlenex:
@JeffVandrewJr I'm aware of the tool, but it's not a newbie-friendly solution, unfortunately.

NicolasDorier:
When Wasabi Wallet and BTCPay are both done with PSBT, I think we will just advise Wasabi Wallet as the recommended wallet @nopara73

nopara73:
# KeyPaths

Note that Wasabi is bech32 only, so that may be a blocker?

# C#

Wasabi is a good choice, because if there's an issue with it, anyone here can easily touch the code, because it's written in C#, too, where your expertise lies.

# Gap Limit

I added a `MinGapLimit`: https://github.com/zkSNACKs/WalletWasabi/commit/409a01cd48fd0753b1c84f1f1ddcef2a93973b9d

The wallet file can be modified from `File/Open/Wallet Folder`.

# Notes

- **This is not a gap limit.** Wasabi may very well generates and monitors more addresses than this if needed, **thus the name minimum gap limit.**  
- **It cannot be less than 21.** If you set it smaller than 21, then Wasabi will autocorrect it to 21.
- **Why 21?** Because there are 21 million bitcoins.
- **Issues can happen at wallet recovery if it's forgotten to be set to the same value again.** So at wallet recovery, don't forget to set the min gap limit again.
[...]

PR #281, opened 2019-07-28
Create ElectrumWallet.md

pavlenex:
This PR create a document that explains how to connect BTCPay and Electrum from creating the new wallet to configuring gap limit.
Addresses #213 

It lacks a proper warning on leaking information to third-party servers, any ideas?


Zaxounette:
An idea of the wording for the warning could be something like this: 

"Be advised that the Electrum wallet rely's on servers that are controlled by third-party individuals.
Informations, like public adresses, amounts transacted can be leaked.
To protect against such leaks, you can setup your own Electrum Personnal Server [LINK TO EPS DOCS]"

pavlenex:
Good suggestion, added a warning now @Zaxounette.

Issue #273, opened 2019-07-23
Add "Can I delete invoices from BTCPay" to FAQ


pavlenex:
Explain why it's not possible to remove the invoices.

PR #291, opened 2019-08-17
Add delete invoices to FAQ

britttttk:
Issue #273

There is probably a more exact reason. @pavlenex do you know?


NicolasDorier:
I think that's good reason. What they can do is use invoice filtering  if they don't want expired to appear.

PR #311, opened 2019-09-26
Add Rasp Pi 4 Guide and update Deployment.md

britttttk:
Add Rasp Pi 4 Guide, thanks @wiz


pavlenex:
@wiz was the idea to have RPI Deployment, then

Raspberry Pi Deployment
- RPI 3.md
- RPI 4.md

So, two separate guides, nested under RPI Deployment section?

I'm thinking how can we have RPI 4 guide focused only on btcpay in docs, I mean, we can't just copy-paste your entire article into RPI4.md :)

PR #326, opened 2019-09-28
Updated the PrestaShop.md

BitcoinMitchell:
The link pointed to a very old version of the PrestaShop plugin.


pavlenex:
@olinkl can you confirm that this is the right link that people should be pointed to?

BitcoinMitchell:
@pavlenex I've assumed it would be, since https://github.com/btcpayserver/prestashop-plugin is a lot more active than the one currently linked. 

I've also opened a few PR's on https://github.com/btcpayserver/prestashop-plugin, so if I got it wrong I have to move those over. 😄

PR #327, opened 2019-10-01
Additions in Deployment and Synchronization FAQ

21isenough:
1. FAQ-Deployment (line 86)
The newly formatted and initiated drive needs to be mounted before the setup process. Otherwise the docker data will be written onto the SD/eMMC card. 

2. FAQ-Synchronization
Rather than deleting the whole '_data' folder, only the 'blocks' and 'chainstate' folders should be deleted and recreated to reindex a node.

### DIFF
--- a/FAQ/FAQ-Deployment.md
+++ b/FAQ/FAQ-Deployment.md
@@ -80,10 +80,11 @@ sudo mkdir /mnt/usb
 UUID="$(sudo blkid -s UUID -o value /dev/sda1)"
 # Add mount to fstab.
 echo "UUID=$UUID /mnt/usb ext4 defaults,nofail 0" | sudo tee -a /etc/fstab
+# Mount the new drive
+mount /dev/sda1
 # Define `/var/lib/docker` as symbolic link to /mnt/usb
 sudo mkdir /mnt/usb/docker
 sudo ln -s /mnt/usb/docker /var/lib/docker
-mount /dev/sda1
 ```
 
 If you want to mount specific folder (like only Bitcoin node data directory), please browse `/var/lib/docker/volumes` to chose the different docker volumes.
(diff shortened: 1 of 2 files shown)


FAQ/FAQ-Deployment.md
```diff
@@ -83,6 +83,7 @@ echo "UUID=$UUID /mnt/usb ext4 defaults,nofail 0" | sudo tee -a /etc/fstab
 # Define `/var/lib/docker` as symbolic link to /mnt/usb
 sudo mkdir /mnt/usb/docker
 sudo ln -s /mnt/usb/docker /var/lib/docker
+mount /dev/sda1
```
Kukks: Doesn't this need to be the first command?

FAQ/FAQ-Synchronization.md
```diff
@@ -125,8 +125,12 @@ To reindex your node:
 
 ```bash
 btcpay-down.sh
-rm -rf /var/lib/docker/volumes/generated_bitcoin_datadir/_data
-mkdir /var/lib/docker/volumes/generated_bitcoin_datadir/_data
+# Delete 'blocks' and 'chainstate' folders
+rm -rf /var/lib/docker/volumes/generated_bitcoin_datadir/_data/blocks
+rm -rf /var/lib/docker/volumes/generated_bitcoin_datadir/_data/chainstate
+# Recreate empty 'blocks' and 'chainstate' folders
+mkdir /var/lib/docker/volumes/generated_bitcoin_datadir/_data/blocks
```
Kukks: I don't think you need to recreate these dirs

FAQ/FAQ-Deployment.md
21isenough: A drive can only be formatted if it's not mounted, so it can't be first. 
I thought I did it in the order above but now that I look at it, I think it should be mounted after the UUID is written to the fstab. Just before creating the 'docker' folder.

FAQ/FAQ-Synchronization.md
21isenough: I think so too, yes. 
I kept it since @NicolasDorier recreated the '_data' folder as well in the commands before. He mentioned something to me about docker having issues with restarting with missing folders.

21isenough: Just checked again: Folders do definitely not need to be recreated. I'll remove these commands and commit it to this PR.

PR #333, opened 2019-10-04
Standardize wallet warrning

### LINKED ISSUE #329: Standard format for warnings
The doc [connectwallet.md](https://docs.btcpayserver.org/getting-started/connectwallet) links to [wallet.md](https://docs.btcpayserver.org/features/wallet) doc, both have a warning with different styling and only a slight variation of text. 

Repeated text is not preferable but since it's a fair warning for users, it would be better if it appeared in a similar format so the reader doesn't have to read it again. 

>Note: A private key (xprv) is never required for receiving money on-chain to a BTCPay wallet. The software needs a public key (xpubkey) which is a watch-only wallet token. The xpubkey allows BTCPay to generate a new address each time a new invoice is generated. It enables users to observe the wallet balance and transactions without having to share their private key. Exception is Lightning Network which is a hot wallet with keys online.

pavlenex:
This PR standardizes the warrning in ConnectWallet.md and Wallet.md
Closes #329

### DIFF
--- a/Wallet.md
+++ b/Wallet.md
@@ -2,7 +2,7 @@
 
 This document explains the features of the internal BTCPay wallet. For frequently asked questions - check the [Wallets FAQ](FAQ/FAQ-Wallet.md)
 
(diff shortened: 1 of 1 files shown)


pavlenex:
Don't like this one, closing

PR #335, opened 2019-10-04
Remove mention of Azure as easiest way to deploy

### LINKED ISSUE #300: Deployment methods - remove Azure
Azure is still displayed in some places and references as the "easiest way to deploy". This should be fixed.

Also, we should be more neutral when recommending a particular deployment method.

pavlenex:
Closes #300 as far as docs are concerned.

### DIFF
--- a/AzureDeployment.md
+++ b/AzureDeployment.md
@@ -4,8 +4,6 @@ This setup is similar to the [Docker Deployment](DockerDeployment.md), except th
 
 ## One-click setup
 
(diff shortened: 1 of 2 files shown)


pavlenex:
Yeah, I think I'll remove that paragraph altogether and re-write. Nicely spotted.

pavlenex:
How it looks now @britttttk ?

pavlenex:
Okay, let me see what I can tackle there, nice list!

PR #347, opened 2019-10-22
Update link

RiccardoMasutti:


pavlenex:
What's the reason behind changing the link, it works as it is?

RiccardoMasutti:
> What's the reason behind changing the link, it works as it is?

New one is updated content, old one is not

pavlenex:
Actually it's not, that's the same link.
ManualDeploymentExtended.md is a relative link. We prefer relative to absolute links where it's possible because of how our content is rendered in GitBook.

RiccardoMasutti:
@pavlenex old link redirects to https://github.com/btcpayserver/btcpayserver-doc/blob/54be4560cfc0a0dbd604d29a9f5948fb21de880a/ManualDeploymentExtended.md
(1 contributor, updated on Jun 1, 1011 lines of code)

New link redirects to https://github.com/btcpayserver/btcpayserver-doc/blob/master/ManualDeploymentExtended.md
(4 contributors, updated on Aug 10, 1036 lines of code)

pavlenex:
As I said, we only use relative links.

britttttk:
Ok I finally understand the issue. 

@RiccardoMasutti on Github it redirects to the correct link. On Gitbook yes you are right, it does not. With docs we need to be clear which one has the docs issue, so we can understand what's happening. :)

britttttk:
Really wondering how that happens @RiccardoMasutti so weird :)

pavlenex:
It's probably the caching issue, will as their support. I only checked on GitBook, probably caching on their end.

pavlenex:
Ok it's fixed now, it was a caching issue, what I did to fix is just did a random commit which triggered gitbook to fetch back that page and render correctly.

Issue #340, opened 2019-10-08
Add "What is a derivation scheme?" to FAQ


britttttk:
Users who create a new invoice without a derivation scheme receive this warning which does not provide information about what it is or how to obtain one.

![warning_noxpub_invoices](https://user-images.githubusercontent.com/39231115/66369481-d10a8700-e959-11e9-9b82-2616ad41c4e4.jpg)

pavlenex:
Done in #355 🚀

PR #355, opened 2019-10-30
Add derivation scheme to FAQ

britttttk:
Probably would also be helpful to give an example maybe when to use something like: xpub...-[p2sh]


britttttk:
#340

Issue #372, opened 2019-11-12
Add additional info to FastSync


britttttk:
Reindex issue: [#1145](https://github.com/btcpayserver/btcpayserver/issues/1145) 
Upadate: [FastSync FAQ](https://docs.btcpayserver.org/faq-and-common-issues/faq-synchronization#can-i-skip-the-synchronization)
Add:

1. Remove existing Bitcoin Core wallet info
```
docker volume rm generated_bitcoin_wallet_datadir
```

2. ~~If any other problems are experienced, also suggest use `echo $BTCPAYGEN_ADDITIONAL_FRAGMENTS` to make sure you have pruning enabled~~ already mentioned in this faq

PR #373, opened 2019-11-13
Update FastSync FAQ

britttttk:
#372 
Not sure if we need to mention the error: `Prune: last wallet synchronisation goes beyond pruned data.`

### DIFF
--- a/FAQ/FAQ-Synchronization.md
+++ b/FAQ/FAQ-Synchronization.md
@@ -31,6 +31,8 @@ btcpay-up.sh
 ```
 After FastSync is complete and you have brought back up your instance, refresh your BTCPay domain and wait for remaining blockchain synchronization. You can also follow [this video](https://youtu.be/VNMnd-dX9Q8?t=1730).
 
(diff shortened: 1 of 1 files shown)


pavlenex:
Mentioning the error would be good, people may google it.

FAQ/FAQ-Synchronization.md
```diff
@@ -31,6 +31,8 @@ btcpay-up.sh
 ```
 After FastSync is complete and you have brought back up your instance, refresh your BTCPay domain and wait for remaining blockchain synchronization. You can also follow [this video](https://youtu.be/VNMnd-dX9Q8?t=1730).
 
+If your FastSync returns: `You need to delete your Bitcoin Core wallet` after you load the uxto set, or you find this error: `Last wallet synchronisation goes beyond pruned data` it means you need to delete your internal Bitcoin Core wallet that is not compatible with your newly pruned node. WARNING: Do not delete this wallet if you have any funds on it. (For example, this may be the case if you use Eclair or FullyNoded to receive funds). If you agree to remove the wallet to finish syncing with FastSync, use `docker volume rm generated_bitcoin_wallet_datadir` before you run `btcpay-up.sh`
```
NicolasDorier: > If your FastSync returns: `You need to delete your Bitcoin Core wallet` after you load the uxto set, or you find this error: `Last wallet synchronisation goes beyond pruned data` it means you need to delete your internal Bitcoin Core wallet that is not compatible with your newly pruned node

If `./load-utxo-set.sh` returns a warning...

britttttk: If it interests you @NicolasDorier that user was saying they ran FastSync with ~10% sync completed

PR #389, opened 2019-12-04
docs(altcoin-faq): add missing word

escapedcat:


escapedcat:
@pavlenex right, forgot about that, thx. done.

Kukks:
I would avoid merging this unless there is a way to allow the old link to exist to route to the new one.

escapedcat:
Hm, it's just an anchor-jump-link. The new one will still point to the correct file at least.  

How are the docs being deployed? Maybe we could add a `div id="how-to-an-altcoin-in-btcpay"` around it here https://github.com/btcpayserver/btcpayserver-doc/pull/389/files#diff-ae48f5bfe6bc26f84e69040465540d98R32 to keep the old link working?

pavlenex:
@Kukks I don't think this is a problem at all.
@escapedcat It's auto-deployed via GitBook.

PR #395, opened 2019-12-13
add how to forward ports with a reverse SSH tunnel

openoms:
Adding the copy of the page from my tutorials which came to be when I was setting up a self-hosted BTCpay on my hack0: https://github.com/openoms/bitcoin-tutorials/blob/master/ssh_tunnel.md

The setup is running non-stop since weeks now without issues.

Any advice to modify, update and link in specific places is welcome.

### DIFF
--- a/ReverseSSHtunnel.md
+++ b/ReverseSSHtunnel.md
@@ -88,4 +88,4 @@ You should be able access the ports/services of the host computer through the IP
 
 https://github.com/rootzoll/raspiblitz/blob/master/FAQ.md#how-to-setup-port-forwarding-with-a-ssh-tunnel
 
-https://stadicus.github.io/RaspiBolt/raspibolt_20_pi.html#login-with-ssh-keys
+https://stadicus.github.io/RaspiBolt/raspibolt_21_security.html#login-with-ssh-keys
--- a/ReverseSSHtunnel.md
+++ b/ReverseSSHtunnel.md
@@ -8,7 +8,7 @@
 ## Requirements:
 * a Virtual Private Server (VPS) - eg. a minimal package on Lunanode for ~3.5$/month
 * root access on the VPS - only root can forward ports under no. 1000
-* ssh access to the host computer (from where the ports will be forwarded from)
+* ssh access to the host computer (where the ports will be forwarded from)
 
 ## On the host computer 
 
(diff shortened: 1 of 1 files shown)


Kukks:
This is invaluable, awesome  stuff

Kukks:
Do you think we could extend this eventually to somehow allow multiple tunnels and forward the ports domain on domain origin? That way we'd be able to offer some kind of service for multiple BTCPay home devices

pavlenex:
@britttttk any idea where we can nest this doc? I'm thinking under changedomain.md, it's more of an advanced doc. Or should it be just put in FAQ instead of having a separate page?

britttttk:
I'm a fan of making it's own doc.
Possibly put under it development section on docs.btcpayserver.org. 

@pavlenex I would avoid adding it to the changedomain.md doc because noobs are confused of that process already. For FAQ I would add a link for reverse SSH tunnel but we can add to FAQ after merge.

pavlenex:
>  Possibly put under it development section on docs.btcpayserver.org.

Or under Raspberry Pi? 
@openoms which section makes more sense to you?

Kukks:
I would do a separate doc. and eventually create a "Accessing your BTCPay installation" which links this, dynamic dns, mDnS(.local domains, there is some trickery here for windows that's undocumented)

openoms:
> Do you think we could extend this eventually to somehow allow multiple tunnels and forward the ports domain on domain origin? That way we'd be able to offer some kind of service for multiple BTCPay home devices

Not sure of what are you thinking of here. The ssh tunnel is running on the BTCPay host, so a VPS could take many tunnels (limited by port numbers) and may be channel them to the specific domain with Nginx?

This is might be what ShiftCrypto is doing for the BitBoxBases called Shift Connect: https://base.shiftcrypto.ch/networking.html

pavlenex:
I just have no clue where this should be link on GitBook so that it does not confuse people. I can merge it as it is, but this doc won't be visible on docs.btcpayserver.org unless you add a link to it in summary.md

openoms:
Could be nested under `deployment`.  There could be an extra layer with advanced deployment options this being one. There could be a page about Tor as well as that topic is not covered in the docs much.

pavlenex:
Ok merging this for now, and will figure out how to link it in GitBook later on.

pavlenex:
Added a new section that now looks like this.
Screenshot 2020-01-03 at 17 24 26

PR #398, opened 2019-12-19
Improve Raspberry Pi 4 docs

dennisreimann:
Most of the tutorial is done as root, so the superfluous `sudo` prefix can be omitted. As the repo is also cloned as root, `cd ~pi/btcpayserver-docker` was not correct.


dennisreimann:
There is the switch to root in the disable swap section – we could male this more prominent though.

Kukks:
Yeah we should move it into  a separate part as the swap section implies it is optional

dennisreimann:
Ok, also did some more polishing while I was at it.

Issue #401, opened 2019-12-26
Add info related to IPN field in app


escapedcat:
Explain to people what teh field is for and what you can expect if you add inpyt email-address into  this field. To cover this: https://github.com/btcpayserver/btcpayserver/issues/1222

NicolasDorier:
@escapedcat the only trusted field of the IPN must be "invoiceId".

Because the callback is not authenticated, you should not try to use any other field. Instead you must extract `invoiceId` then fetch the invoice. The other fields are here purely for backward compatibility reason with Bitpay's API.

escapedcat:
Uhm, I am not using this. I wanted to document it because I thought I might want to use it but after [@pavlenex explaination](https://github.com/btcpayserver/btcpayserver/issues/1222#issuecomment-564240989) I didn't want to anymore.  
With the PR I wanted to make clear what this is and what you can expect from it.  
I can add your info to it to make it more clear.  

I now feel this field could also be removed if the purpose or usage is unclear or misleading. @pavlenex [mentioned something similar in his reply](https://github.com/btcpayserver/btcpayserver/issues/1222#issuecomment-564240989).

NicolasDorier:
@escapedcat my apologies I read too fast. I thought you were refering to a field in the json of the IPN. You are right, it is not clear.

escapedcat:
No worries. So will this be there and the documentation maybe helps to explain the prupose or is there a chance that the field will be removed soonish? I can close this. I'm fine with either. Just want to avoid that people might also wonder what this field is about like I did.

NicolasDorier:
I am fine with either, I don't really understand the intention of this field myself. @pavlenex @Kukks

Kukks:
I ack for removing it. It is such a weird feature to have in 2019/2020

pavlenex:
Ok, since we have 4 ACKs for removing this feature, it would be good to close this issue and open a request in the main repository for removing this option from both apps.

PR #402, opened 2019-12-26
feat(faq-apps): add ipn info #401

escapedcat:
This could then be linked with a `?`-icon from the related field.


pavlenex:
Considering [we'll remove this most likely, ](https://github.com/btcpayserver/btcpayserver/issues/1241)I'd close this PR if that's okay with you.

escapedcat:
Yes, makes sense.

Issue #404, opened 2019-12-30
Decision chart for deployment


NicolasDorier:
I am creating a decision chart and I am waiting for some feedback.

![image](https://user-images.githubusercontent.com/3020646/71568614-0a433500-2b0c-11ea-9b46-d563b726cd41.png)

I will apply the lipstick once the content is good enough. @britttttk @pavlenex

NicolasDorier:
There is other alternative to nodlit, but I can't remember names

britttttk:
This is great. It's really good for explaining choices.

NicolasDorier:
This will be added to this page.

https://github.com/btcpayserver/btcpayserver-doc/blob/73ef16b3c916ad8ff9a2d7dd3a60404b56cc48a7/Deployment.md#btcpay-server-deployment

NicolasDorier:
v2

![image](https://user-images.githubusercontent.com/3020646/71569698-509c9200-2b14-11ea-83e5-07bc4e600ab4.png)

britttttk:
ACK

NicolasDorier:
Final version:

![DecisionDiagInstallBTCPayServer](https://user-images.githubusercontent.com/3020646/71572461-98c2b100-2b22-11ea-9278-5a830604d71c.png)

pavlenex:
Nice,I would avoid advertising particular hosts (LunaNode, Hack0, etc) and rather point them out to generic term "Web Deployment", because in future we may have more of these and then it'll be confusing for people.

I also am not sure about first step. First step should point to third-party hosting and then to self-host.

NicolasDorier:
@pavlenex there is no generic Web deployment right now, there is no alternative to LunaNode.

NicolasDorier:
I think we can rewrite those when we actually have generic web deployment and more boxes.

pavlenex:
Well if you are wiling to update it, then okay. Make sure to keep the file in the format that is easy to edit later on.

NicolasDorier:
yeah I commited the visio file

PR #405, opened 2019-12-30
Add a decision diagram to explain deployment possibilities

NicolasDorier:
Close https://github.com/btcpayserver/btcpayserver-doc/issues/404

DecisionDiagInstallBTCPayServer


NicolasDorier:
Fixed the "strenghen" => "strenghens"

NicolasDorier:
@britttttk nop that was visio which exported only the element I was selecting, good catch :p
Should be good now.

PR #415, opened 2020-01-18
Add confirmed invoice status

britttttk:
Add confirmed invoice status to invoices page for issue [#1288](https://github.com/btcpayserver/btcpayserver/issues/1288)

### DIFF
--- a/Invoices.md
+++ b/Invoices.md
@@ -23,7 +23,7 @@ Table below lists and describes common invoice statuses in BTCPay and suggests c
 | New                    | Not paid, invoice timer still has not expired                | None                                                         |
 | Expired                | Not paid, invoice timer expired                              | None                                                         |
 | Paid                   | Paid, but has not received sufficient amount of confirmations specified in the Store Settings | Wait for confirmations (The invoice should become - complete) |
-| Confirmed                   | Paid, confirmed, by reaching the number of confirmations in store settings but has not received the default number (6) of confirmations in BTCPay. (This status is not used for Lightning Network transactions) | Wait for remaining confirmations, or proceed if store settings confirmations is acceptable  |
+| Confirmed***                   | Paid, confirmed, by reaching the number of confirmations in store settings but has not received the default number (6) of confirmations in BTCPay. | Wait for remaining confirmations, or proceed if store settings confirmations is acceptable  |
 | Completed              | Paid, completed, received sufficient amount of confirmations in store | Fulfil the order                                             |
 | Expired (paidPartial)* | Paid, but not in full amount and expired                     | Contact buyer to arrange a refund or ask for them to pay their due. Optionally mark invoice as completed |
 | Paid (Paid Over)       | Paid more than the invoice amount, but has not received sufficient amount of confirmations specified in the Store Settings | Wait for confirmations. The invoice should become - Complete (Paid Over) |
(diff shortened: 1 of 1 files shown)


Transisto:
> Paid, confirmed, by reaching the number of confirmations in store settings but has not received the default number (6) of confirmations in BTCPay.

Where can BTCPAY default of 6 confirmation can be changed?

6 conf. is arbitrary

britttttk:
It’s hard coded value in btcpay. @Transisto

bolatovumar:
@Transisto you can change this setting in your store settings I believe.

britttttk:
I think I may have caused some confusion by calling this value "default number (6) of conf in btcpay" for simplicity, but I can explain more with an example. 

I have store settings set to: 1 Confirmation

In the invoice page example below there are some statuses shown:
1. Paid - Tx has been received by the network and is in the mempool (0 conf in this example)
2. Confirmed - Has at least 1 confirmation (store settings, externally set)
3. Complete - Has received the max tracked confirmations (6, [internally set](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer.Common/BTCPayNetwork.cs#L62))
4. New - Zero payments received yet

image

Once the tx has received the max tracked confirmations, it is no longer monitored by InvoiceWatcher.
So it's not really a default conf value but more like when btcpay stops watching that invoice's status. 

image

@bolatovumar @Transisto So I'm not sure maybe it should be called something else in the table instead of default conf maybe like stop monitoring conf value or something. But you would need to modify the source code to increase it.

Transisto:
5, 6 or 7 confirmation is arbitrary and mean actually nothing to merchants,
therefore having a status change name and color in the UI because of some
backend process is kinda seeking confusion and support requests.

If anything the hardcoded 6 should be jacked up so merchant of high risk
items can set it higher.

bolatovumar:
@Transisto it's only "arbitrary and mean actually nothing to merchants" if the merchant knows nothing about Bitcoin. We assume here that if the merchant is using BTCPay then the merchant knows something about Bitcoin. Having six confirmations on a transaction before it's considered "safe" is a very common practice in Bitcoin which most merchants probably already know about.

I guess one thing that could be done here is allowing merchants to pick how many confirmations they want before the transaction is considered "complete". It would be set to 6 by default (current value) but merchants could update it to any number more than or equal to 1.

Transisto:
I'm aware of the old 6 confirmation meme and think it's sad that it pass as
a substitute for people doing their own risk management. I think a cap of 6
is sufficient but when 3 conf is selected by merchant it's causing
confusion to display an extra status with its own color change. It's not
useful for a merchant to know that the 6 conf hardcoded limitation of the
back end has been reached.

bolatovumar:
@Transisto are you suggesting this additional "six-conf complete" status be removed entirely then?

I can understand that this additional status change may be confusing to some. I guess there are two options here:

* Remove this additional status -- this fixes the issue of potential confusion as there will be no additional status change and therefore nothing that could confuse the merchant

OR

* Explain to the user in the UI what this status is through a tooltip or something of the sort -- this will help reduce confusion by providing an explanation. Also this has an additional benefit of not changing the UI for those merchants who are used to the existing behavior

britttttk:
>Explain to the user in the UI what this status is through a tooltip

Tooltip currently links to [accounting](https://docs.btcpayserver.org/features/accounting). As @bolatovumar mentioned, it would probably be a good idea either way, to add a link to the invoices doc on that page to lead readers to status explanations.

As for the confirmed/completed status change, I suggest to open issue in the main btcpayserver repo to get some more input on why `max tracked confirmations` is used and what it would require to change it.

bolatovumar:
@Transisto it's probably a good idea to open a new issue in the main repo about this as @britttttk mentioned to get some further input on what should be done here if anything

PR #417, opened 2020-01-23
Add Vault.md

### LINKED ISSUE #409: Document BTCPay Vault and HWI in "Connect a wallet"
Connect a wallet should have a BTCPay Vault (or Hardware Wallet) section that will explain step by step how to connect and spend from supported hardware wallets through the Vault.

pavlenex:
Closes #409 
Vault right now is just a separate page, since we'll be doing some UI tweaks in the future, I've yet to think of an appropriate place for this document, since it would require the re-structure of the connect a wallet section and internal wallet section for which I'll open a new issue. 

For now, please review the text for clarity/grammar.


britttttk:
Only comment I have is, that it would be good to somehow indicate to the reader that they do not have to set up their wallet with Vault to spend using Vault. Meaning, users who already have a wallet connected can skip to Step 2 of the article or sth..

pavlenex:
I'm not sure about that one @britttttk. I'd rather them set up a from scratch properly then having to deal with address formats and stuff. With Vault, you have a dropdown menu, select appropriate format which is displayed in a user-friendly way and that's it. I'm afraid people will mess things up if we advise them to do it manually.

PR #426, opened 2020-01-29
Added Link to hardware config

Dayrook:
Feel free to reword and make the change however you want - just trying to save other people a few days of frustration!
----
Existing procedure will not function due to hardware not playing nicely with OS. At least 1 kernel parameter, `intel_idle.max_cstate=1`, and most likely some BIOS changes need to be made.


Dayrook:
It's the exact device recommended earlier in this hardware deployment
guide, which means following step-by-step will get you a locked up system.

pavlenex:
Yeah, that what I was thinking the tutorial actually recommends that single device, perhaps in the future we can re-write to make it less specific @Kukks, but for now LGTM?

Kukks:
https://media.giphy.com/media/BHKpXZnLkYFYk/giphy.gif

PR #435, opened 2020-02-10
Document the workaround for issue with 503 error

alevchuk:
See workaround here
https://github.com/btcpayserver/btcpayserver-docker/issues/263#issuecomment-581472247

The "set -u; echo "$BTCPAY_HOST" makes sure the variable is set. If not, it will print
> -bash: BTCPAY_HOST: unbound variable
and exit early.

### DIFF
--- a/FAQ/FAQ-Deployment.md
+++ b/FAQ/FAQ-Deployment.md
@@ -379,3 +379,28 @@ You will also need those settings in the `apache2.conf` to prevent issues while 
 LimitRequestLine 500000
 LimitRequestFieldSize 500000
 ```
+
+#### When I try to access my BTCPay by IP address I get 503 Service Temporarily Unavailable nginx
+
+Your nginx config is set to route the HTTP request to a particular container based on the domain name of the request. For example, the official [deployment on pi 4](https://docs.btcpayserver.org/deployment/raspberrypideployment/rpi4) was to setup the souce domain name to http://raspberrypi.local/ yet getting automatic local domain raspberrypi.local does not always work. You are problably in this situation and trying to type the IP address of your BTCPay into the web-browser.
+
+Since nginx does not sees the IP address in the request instead of raspberrypi.local it does not know where to route that reuqstes and returns:
+```
+503 Service Temporarily Unavailable
+-----------------------------------
+nginx
+```
+
+You can fix this by forcing nginx to route the HTTP request to BTCPay even if request  domain name is not recognize.
+Simply, re-run the setup script like this:
+
+```
+sudo su -
+
+(set -u; echo "$BTCPAY_HOST") && REVERSEPROXY_DEFAULT_HOST="$BTCPAY_HOST" && . btcpay-setup.sh -i
+
+```
+
+Now putting local IP in the web-browser works.
+
+(Source: https://github.com/btcpayserver/btcpayserver-docker/issues/263#issuecomment-581472247 )
--- a/FAQ/FAQ-Deployment.md
+++ b/FAQ/FAQ-Deployment.md
@@ -384,22 +384,23 @@ LimitRequestFieldSize 500000
 
 #### Cause 1: Trying to access my BTCPay by IP address
 
-Your nginx config is set to route the HTTP request to a particular container based on the domain name of the request. For example, the official [deployment on pi 4](https://docs.btcpayserver.org/deployment/raspberrypideployment/rpi4) was to setup the souce domain name to http://raspberrypi.local/ yet getting automatic local domain raspberrypi.local does not always work. You are problably in this situation and trying to type the IP address of your BTCPay into the web-browser.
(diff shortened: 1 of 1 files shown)


pavlenex:
I'm unsure if this is the only situation in which 503 happens, so @NicolasDorier it's up to you how you want this handled.

pavlenex:
After thinking a bit more about this, 503 is such a common error that it would be nice if this could be a beginning of something similar to [troubleshooting sync doc,](https://docs.btcpayserver.org/faq-and-common-issues/faq-synchronization#btcpay-server-takes-forever-to-synchronize) with cases 1, case 2, case 3, etc.

If we just let it in the wild like this people are going to be confused, since this is not the only situation in which 503 happens.

FAQ/FAQ-Deployment.md
```diff
@@ -379,3 +379,28 @@ You will also need those settings in the `apache2.conf` to prevent issues while
...
+```
+sudo su -
+
+(set -u; echo "$BTCPAY_HOST") && REVERSEPROXY_DEFAULT_HOST="$BTCPAY_HOST" && . btcpay-setup.sh -i
+
+```
+
+Now putting local IP in the web-browser works.
+
+(Source: https://github.com/btcpayserver/btcpayserver-docker/issues/263#issuecomment-581472247 )
```
pavlenex: The source part is redundant imo.

FAQ/FAQ-Deployment.md
```diff
@@ -379,3 +379,54 @@ You will also need those settings in the `apache2.conf` to prevent issues while
...
 LimitRequestFieldSize 500000
 ```
+
+### I get 503 Service Temporarily Unavailable nginx
+
+#### Cause 1: Trying to access my BTCPay by IP address
+
+Your nginx config is set to route the HTTP request to a particular container based on the domain name of the request. For example, the official [deployment on pi 4](https://docs.btcpayserver.org/deployment/raspberrypideployment/rpi4) was to setup the souce domain name to http://raspberrypi.local/ yet getting automatic local domain raspberrypi.local does not always work. You are problably in this situation and trying to type the IP address of your BTCPay into the web-browser.
+
+Since nginx does not sees the IP address in the request instead of raspberrypi.local it does not know where to route that reuqstes and returns:
```
NicolasDorier: `Since nginx does see the IP address`

NicolasDorier: `reuqstes` typo

FAQ/FAQ-Deployment.md
```diff
@@ -379,3 +379,54 @@ You will also need those settings in the `apache2.conf` to prevent issues while
...
+```
+503 Service Temporarily Unavailable
+-----------------------------------
+nginx
+```
+
+You can fix this by forcing nginx to route the HTTP request to BTCPay even if request  domain name is not recognize.
+Simply, re-run the setup script like this:
+
+```
```
NicolasDorier: add `bash` at the end of the back tilts so we get better syntax colorization

FAQ/FAQ-Deployment.md
```diff
@@ -379,3 +379,54 @@ You will also need those settings in the `apache2.conf` to prevent issues while
...
+(set -u; echo "$BTCPAY_HOST") && REVERSEPROXY_DEFAULT_HOST="$BTCPAY_HOST" && . btcpay-setup.sh -i
+
+```
+
+Now putting local IP in the web-browser works.
+
+#### Cause 2: btcpayserver or letsencrypt-nginx-proxy is not running
+
+To check, run: 
+```
```
NicolasDorier: add bash at the end of the back tilts so we get better syntax colorization

FAQ/FAQ-Deployment.md
```diff
@@ -379,3 +379,54 @@ You will also need those settings in the `apache2.conf` to prevent issues while
...
+nginx
+```
+
+You can fix this by forcing nginx to route the HTTP request to BTCPay even if request  domain name is not recognize.
+Simply, re-run the setup script like this:
+
+```
+sudo su -
+
+(set -u; echo "$BTCPAY_HOST") && REVERSEPROXY_DEFAULT_HOST="$BTCPAY_HOST" && . btcpay-setup.sh -i
```
NicolasDorier: This is weird you need that...

NicolasDorier:
Remove the hack on `set -u`. If `BTCPAY_HOST` is not set, you have a bigger problem. `BTCPAY_HOST` MUST be set before setting `REVERSEPROXY_DEFAULT_HOST`

FAQ/FAQ-Deployment.md
alevchuk: it's in case folks don't have BTCPAY_HOST set, `set -u` fails early and says exactly what's wrong

alevchuk:
> Remove the hack on `set -u`. If `BTCPAY_HOST` is not set, you have a bigger problem. `BTCPAY_HOST` MUST be set before setting `REVERSEPROXY_DEFAULT_HOST`

Oh, i see

pavlenex:
Thanks for those changes @alevchuk I have a few more ideas on other cases that I'd be happy to document once we merge this one.

Waiting for @NicolasDorier to do a final review and merge if it passes his review, on my end it looks good.

PR #436, opened 2020-02-10
Prevent people from running into issue #263

alevchuk:
This implements [the workaround](https://github.com/btcpayserver/btcpayserver-docker/issues/263#issuecomment-581472247) for new deployments


NicolasDorier:
Good idea

PR #468, opened 2020-03-26
Change dropdown view to link to css

artdesignbySF:
Temporarily change the "click to view" dropdown css to linking to a dedicated page per CSS as long as dropdown markdown isn't supported.


pavlenex:
With doc migration this is no longer an issue! 🎸

artdesignbySF:
Oh wow, it looks sleek! Nice! :D

Issue #486, opened 2020-04-17
Queries on possible changes to payjoin-spec


AdamISZ:
The following refers to [this](https://github.com/btcpayserver/btcpayserver-doc/blob/eaac676866a4d871eda5fd7752b91b88fdf849ff/Payjoin-spec.md) version of the doc.

First a thanks to @Kukks for bringing this to my attention, and to the whole project for implementing this, it looks like a major step forward.
I wanted to review a little bit by comparing with what I did in Joinmarket a year or two back. Also apologies for not reviewing the code itself, but that would be very difficult for me.

Here's what I came up with:

* The spec does not mention input/output ordering. This is mentioned in the original BIP79 but I think could be clearer. I can see an argument for just enforcing BIP69 everywhere, as it makes it easy for both sides to reason about.
* I already mentioned this in earlier discussions, and people don't seem to agree with me, but just for the record: in JM i have the sender specify a (payjoin) version range and the receiver responds with a choice from within that range (in JM only a version 01 exists for now). Although one can retrofit versioning (no version = version 0 so to speak), I don't see an advantage in not being explicit.
* You don't mention the UIH1 or UIH2 heuristics that were discussed [here](https://gist.github.com/AdamISZ/4551b947789d3216bacfcb7af25e029e#gistcomment-2799709). Does that mean the receiver utxo selection doesn't pay attention to that? To be clear, I don't think that's necessarily a problem, but one can argue it's useful to address it. In the JM implementation the approach can be summarized "try to avoid UIH2 but go ahead anyway even if you can't". I guess if you *had* included this in the code it would have been mentioned [here](https://github.com/btcpayserver/btcpayserver-doc/blob/eaac676866a4d871eda5fd7752b91b88fdf849ff/Payjoin-spec.md#blockchain-analytics-heuristics-affected-by-our-implementation-).
* "Then the original PSBT is deemed valid, however, other checks are done which depends on the state of the receiver's state" <-- typo
* "Check that the sender's input's sequence numbers are unchanged." <-- this wording does not suggest that the new inputs from the receiver are checked to have the same sequence number, also (which they should be).
* [Typo](https://github.com/btcpayserver/btcpayserver-doc/blob/eaac676866a4d871eda5fd7752b91b88fdf849ff/Payjoin-spec.md#rational), should be "Rationale".
[...]

PR #487, opened 2020-04-18
Additional note on input/output ordering

NicolasDorier:


NicolasDorier:
About https://github.com/btcpayserver/btcpayserver-doc/issues/486 @AdamISZ, does this make sense and is better explained?

AdamISZ:
Yes, very much ACK. And I appreciate you even thought a step further!

I think this perfectly illustrates something quite deep: the bitcoin protocol *mostly does not* and *entirely should not*, assign any meaning to transaction input, output ordering. I say "mostly" because of weird things around sighash_single.

Consequently it would be much better, I now believe, that BIP69 did not exist, and that people never add any meaning to the order of inputs and outputs. If their code constructs transactions in non-random order then that adds meaning to ordering, so that should always be considered a bug in wallet code.

As for now, the reality is that there is some small-ish subset of transactions that are indeed using BIP69, so what you describe here could make sense. Unfortunately I do still see a slight problem: "if the sender uses BIP69" - you can't tell that for sure though? So if a proposal from the sender is BIP69 conforming, that may have occurred by chance (indeed, it does happen by chance in a very non-trivial fraction of randomly created txs). If you then make the payjoin BIP69 then the distribution of their payments which are BIP69 conforming will be slightly different for their payjoins compared with their non-payjoins.

NicolasDorier:
> Unfortunately I do still see a slight problem: "if the sender uses BIP69" - you can't tell that for sure though?

I think that's fine. The payjoin transaction could also be accidentally BIP69 compliant.
Certainly, it means that the distribution of BIP69 payments would be slightly higher than non-bip69 one, but because of the usage of PJ, such distribution is already hard to know.

Do you know any wallet using BIP69 btw? If there is no user facing wallet using it, it is maybe better ignoring this for now.

Issue #478, opened 2020-04-12
Replace old design images


pavlenex:
Legacy theme images should be replaced with new light theme images. Especially in walkthrough, but generally everywhere. Also new invoice images should be replaced.

PR #490, opened 2020-04-21
Image replace

pavlenex:
Replaced a few images on important places, did not go way too crazy as some images require labeling and editing. Replaced only the places where a newbie may see a different interface and freak out.


pavlenex:
Closes #478 unless someone wants to go crazy and replace images everywhere, which is tedious and unnecessary at this point imo

PR #493, opened 2020-04-27
Add payjoin risks

NicolasDorier:
We were not properly documenting the risks inherent to payjoin.

### DIFF
--- a/Payjoin-spec.md
+++ b/Payjoin-spec.md
@@ -139,7 +139,7 @@ We then send the payjoin transaction proposal back to the sender.
 
 \*\* <a name="receiver-pay"></a>: If the sender overpays the invoice, we can substract fees from his over paid amount. This is especially useful when somebody wants to make a donation from a small unspent output in his own wallet the amount of his change may be overpaying the invoice. If we were not substracting the fees on the output paying the invoice, we would fall into a `not-enough-money` error, as no change exists.
 
-### Sender side
+### Sender side <a name="sender"></a>
 
 Before sending the original PSBT to the receiver, the sender should make sure that the original PSBT is at least attempted to be broadcasted automatically after 1 minute of submission.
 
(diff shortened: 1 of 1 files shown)


Payjoin-spec.md
```diff
@@ -222,6 +222,34 @@ If Alice pays Bob, she might be tempted to pay him a round amount, like `1.23000
 
 For this reason, during a [spare change](#spare-change) situation, we randomly round the amount in the output added by the receiver to the payjoin transaction.
 
+## Risks
+
+### On the receiver side: UTXO probing attack
+
+When the receiver creates a payjoin proposal, he exposes one or several inputs belonging to him.
```
Kukks, line 229: `they expose one or more inputs belonging to them.`

Payjoin-spec.md
```diff
@@ -222,6 +222,34 @@ If Alice pays Bob, she might be tempted to pay him a round amount, like `1.23000
...
+
+An attacker could create lot's of original transaction in order to learn the UTXOs of the receiver, while not broadcasting the payjoin proposal.
```
Kukks, line 231: `An attacker could create multiple original transactions in order to learn the UTXOs of the receiver, while not broadcasting the payjoin proposal.`

Payjoin-spec.md
```diff
@@ -222,6 +222,34 @@ If Alice pays Bob, she might be tempted to pay him a round amount, like `1.23000
...
+
+We cannot fully prevent this type of attack, we implemented the following mitigations:
```
Kukks, line 233: ` While we cannot prevent this type of attack entirely, we implemented the following mitigations:`

Payjoin-spec.md
```diff
@@ -222,6 +222,34 @@ If Alice pays Bob, she might be tempted to pay him a round amount, like `1.23000
...
+
+* When the receiver detects an original transaction being broadcasted, or if the receiver detect that the original transaction has been double spent, then he will reuse the UTXO he exposed for the next payjoin.
```
Kukks, line 235: `When the receiver detects an original transaction being broadcasted, or if the receiver detects that the original transaction has been double spent, then they will reuse the UTXO that was exposed for the next payjoin.`

Payjoin-spec.md
```diff
@@ -222,6 +222,34 @@ If Alice pays Bob, she might be tempted to pay him a round amount, like `1.23000
...
+* Only a single payjoin is possible per invoice.
+* While the exposed UTXO will be reused in priority to not leak other UTXOs, there is no strong guarantee about it. This prevent the attacker from detecting with certainty the next payjoin of the merchant to another peer.
```
Kukks, line 237: `While the exposed UTXO will be reused in priority to not leak other UTXOs, there is no strong guarantee about it. This prevents the attacker from detecting with certainty the next payjoin of the merchant to another peer.`

Payjoin-spec.md
```diff
@@ -222,6 +222,34 @@ If Alice pays Bob, she might be tempted to pay him a round amount, like `1.23000
...
+
+Note that probing attack is only a problem for automated payment system like BTCPay Server. End users wallet with payjoin capabilities are not affected, as the attacker can't create multiple invoice to force the receiver to expose his UTXOs.
```
Kukks, line 239: `Note that probing attacks are only a problem for automated payment systems such as BTCPay Server. End-user wallets with payjoin capabilities are not affected, as the attacker can't create multiple invoices to force the receiver to expose their UTXOs.`

Payjoin-spec.md
```diff
@@ -222,6 +222,34 @@ If Alice pays Bob, she might be tempted to pay him a round amount, like `1.23000
...
+
+### On the sender side: Double payment risk for hardware wallets
+
+For a successfull payjoin to happen the sender need to sign two transactions double spending each other: The original transaction and the payjoin proposal.
```
Kukks, line 243: `For a successful payjoin to happen, the sender needs to sign two transactions double spending each other: The original transaction and the payjoin proposal.`

Payjoin-spec.md
```diff
@@ -222,6 +222,34 @@ If Alice pays Bob, she might be tempted to pay him a round amount, like `1.23000
...
+
+BTCPay Server or end user wallet can verify that the payjoin proposal is legit by following [the sender's steps explained earlier](#sender).
```
Kukks, line 245: `BTCPay Server or end user wallets can verify that the payjoin proposal is legitimate by following [the sender's steps explained earlier](#sender).`

Payjoin-spec.md
```diff
@@ -222,6 +222,34 @@ If Alice pays Bob, she might be tempted to pay him a round amount, like `1.23000
...
+
+However, a hardware wallet can't verify that this is indeed the case. This mean that the security guarantee of the hardware wallet decreased. If BTCPay Server or the end user wallet were compromised, the hardware wallet would sign two valid transactions, thus sending two payments.
```
Kukks, line 247: `However, a hardware wallet can't verify that this is indeed the case. This means that the security guarantee of the hardware wallet is decreased. If BTCPay Server or the end user wallet is compromised, the hardware wallet would sign two valid transactions, thus sending two payments.`

PR #497, opened 2020-05-01
Add "Hardening your VPS" chapter

RiccardoMasutti:
This should give some ideas on how to secure your VPS.

### DIFF
--- a/DockerDeployment.md
+++ b/DockerDeployment.md
@@ -16,3 +16,73 @@ This method of deployment is used under the hood for:
 * [LunaNode Web Deployment](LunaNodeWebDeployment.md) *
 * [Azure Deployment](AzureDeployment.md) *
 * [Google Cloud Deployment](GoogleCloudDeployment.md) *
+
+## Hardening your Linux VPS
+
+Securing your BTCPay Server istance is important to protect your and your customers' data from the hands of hackers.
+It would take a whole chapter to explain best security practices; for this reason, we will simply explain the basic hardening.
+If you want to deepen this topic, you can follow [this tutorial](https://www.cyberciti.biz/tips/linux-security.html).
+
+###  Use strong passwords
+
+Make sure you set a good and strong password for your users.
+A good password should be at least 12 characters long and a mixture of alphabets, number, special character, upper and lower alphabets, etc.
+
+### Harden the BIOS
+
+Enable BIOS password and protect GRUB with password to restrict physical access of your system.
+Next, disable booting from CD/DVD, External Devices, Floppy Drive in BIOS.
+
+Steps:
+* Open a terminal and type `grub-md5-crypt`.
+* Digit your desired password and retype it again to confirm it.
+Output should be:
+```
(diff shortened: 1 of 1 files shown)


DockerDeployment.md
```diff
@@ -16,3 +16,73 @@ This method of deployment is used under the hood for:
...
 * [Azure Deployment](AzureDeployment.md) *
 * [Google Cloud Deployment](GoogleCloudDeployment.md) *
+
+## Hardening your Linux VPS
+
+Securing your BTCPay Server istance is important to protect your and your customers' data from the hands of hackers.
+It would take a whole chapter to explain best security practices; for this reason, we will simply explain the basic hardening.
+If you want to deepen this topic, you can follow [this tutorial](https://www.cyberciti.biz/tips/linux-security.html).
+
+###  Use strong passwords
```
pavlenex, line 26: Better suggest here to use a [ssh-key-verifcation](https://www.digitalocean.com/community/tutorials/how-to-configure-ssh-key-based-authentication-on-a-linux-server) since this is something advanced users will be performing.

RiccardoMasutti: Initially I had included the guide on how to disable password access and instead use keys.
I can reinsert it again if you want

pavlenex: I think keys are much better suggestions, considering the section is called "hardening".

RiccardoMasutti: If on the one hand safety increases, on the other it increases risks.

I think that for the average user this information is enough, authentication by keys I see it more suitable for more experienced users (who already know these things).

Do not you think?

pavlenex: How about we just link to it and mentioned it as a good practice?

pavlenex:
@RiccardoMasutti  Since we no longer have ` DockerDeployment.md` if you still want this in, please add it in FAQ-Deployment and FAQ-Readme.

pavlenex:
@RiccardoMasutti If you still want this in, would be cool if you can resolve the conflicts when you get the time.

DockerDeployment.md
```diff
@@ -16,3 +16,73 @@ This method of deployment is used under the hood for:
 * [LunaNode Web Deployment](LunaNodeWebDeployment.md) *
 * [Azure Deployment](AzureDeployment.md) *
 * [Google Cloud Deployment](GoogleCloudDeployment.md) *
+
+## Hardening your Linux VPS
+
+Securing your BTCPay Server istance is important to protect your and your customers' data from the hands of hackers.
```
Zaxounette, line 22:
```suggestion
Securing your BTCPay Server instance is important to protect your and your customers' data from the hands of hackers.
```

Eskyee:
SSH Key-Based Authentication is great for advance users, but  for noobs can be risky as if they lose their SSH Key-Based Authentication they also lose  their  access ... ie like wiping & updating their home pc and not keeping a copy of their SSH Key.

just a warning for noobs !! @pavlenex

PR #505, opened 2020-05-04
Add sender optional parameters for payjoin, and information about error format

NicolasDorier:
Those are not breaking any existing implementation, but I think those are nice to have.

### DIFF
--- a/Payjoin-spec.md
+++ b/Payjoin-spec.md
@@ -74,12 +74,19 @@ However, we also accept the PSBT or a bitcoin transaction in hex format.
 
 When the payjoin sender sends the POST request to the payjoin endpoint, he can optionally specify the following parameters:
 
-* `v=`, the version number of the payjoin protocol that the client is using. The current version is `1`.
+* `v=`, the version number of the payjoin protocol that the sender is using. The current version is `1`.
 
-This can be used in the future so the receiver can reject a payjoin if the client is using a version which is not supported via an error HTTP 400, `version-unsupported`.
-If not specified, the receiver will assume the sender is using the same version.
+This can be used in the future so the receiver can reject a payjoin if the sender is using a version which is not supported via an error HTTP 400, `version-unsupported`.
+If not specified, the receiver will assume the sender is `v=1`.
 
-If the version of the sender is above the one of the receiver, the receiver should ignore the parameter.
+If the receiver does not support the version of the sender, it should send an error with the list of supported versions:
+```json
+{
+    "errorCode": "version-unsupported",
+    "supported" : [ 2, 3, 4 ],
+    "message": "The version is not supported"
+}
+```
 
 * `feebumpindex=`, the preferred output from which to increase the fee for the added inputs. (default: `-1`)
 
@@ -95,8 +102,8 @@ The receiver may send errors to the client, the format is the following:
 
 ```json
 {
-    "errorCode": "version-unsupported",
-    "message": "The version is not supported anymore"
+    "errorCode": "psbt-not-finalized",
+    "message": "You should finalize the PSBT before giving it to the receiver"
 }
 ```
 
--- a/Payjoin-spec.md
+++ b/Payjoin-spec.md
@@ -72,7 +72,7 @@ However, we also accept the PSBT or a bitcoin transaction in hex format.
 
 ### Optional parameters
 
-When the payjoin sender sends the POST request to the payjoin endpoint, he can optionally specifies the following parameters:
+When the payjoin sender sends the POST request to the payjoin endpoint, he can optionally specify the following parameters:
 
 * `v=`, the version number of the payjoin protocol that the client is using. The current version is `1`.
 
--- a/Payjoin-spec.md
+++ b/Payjoin-spec.md
@@ -70,6 +70,42 @@ There is no guarantee concerning the ordering of outputs and inputs in either th
 Note: We strongly recommend the sender to submit a PSBT in Base64 format.
 However, we also accept the PSBT or a bitcoin transaction in hex format.
 
+### Optional parameters
+
+When the payjoin sender sends the POST request to the payjoin endpoint, he can optionally specifies the following parameters:
+
+* `v=`, the version number of the payjoin protocol that the client is using. The current version is `1`.
+
+This can be used in the future so the receiver can reject a payjoin if the client is using a version which is not supported via an error HTTP 400, `version-unsupported`.
+If not specified, the receiver will assume the sender is using the same version.
+
+If the version of the sender is above the one of the receiver, the receiver should ignore the parameter.
+
+* `feebumpindex=`, the preferred output from which to increase the fee for the added inputs. (default: `-1`)
+
+If the `feebumpindex` is out of bounds or pointing to the payment ouptut meant for the receiver, the receiver should ignore the parameter.
+
+* `noadjustfee=`, a boolean the sender specifies if they prefer that the receiver does not bump the fee for the added input. (default: `false`)
+
+Note that if `noadjustfee` is `true`, the sender should create a transaction with RBF disabled, as the original transaction could replace the payjoin transaction.
+
+### Error format
+
+The receiver may send errors to the client, the format is the following:
+
+```json
+{
+    "errorCode": "version-unsupported",
+    "message": "The version is not supported anymore"
+}
+```
+
+It is recommended for the sender software to show a properly formatted message from the `errorCode` which are detailed below.
+
+If the sender software does not recognize an `errorCode`, it should show a generic error to the user.
+
+Unknown `errorCode` and `message` should never be shown in the sender's software UI as it could be used by a malicious receiver to mislead a user. However, they could be shown in the debug logs.
+
 ## Rationale
 
 Here is the rationale for using PSBT instead of raw transactions, and why the receiver should be responsible to bump the fee of the payjoin transaction.


NicolasDorier:
Ping @Kukks , we should also probably ping wasabi, coinjoin and bluewallet to get feedback.

Kukks:
cc @junderw @lukechilds @lontivero @AdamISZ @ncoelho

junderw:
So the response body is:
A hex PSBT
B base64 PSBT
C json object with error.

Why not have it all be json?

```js
// success
{
  "success": true,
  "format": "hex", // or "base64"
  "type": "psbt", // or "rawtx"
  "data": "xxxxxx(hex psbt)"
}
// failure
{
  "success": false,
  "httpCode": 400,
  "errorCode": "version-unsupported",
  "message": "The version is not supported anymore"
}
```

Payjoin-spec.md
```diff
@@ -70,6 +70,43 @@ There is no guarantee concerning the ordering of outputs and inputs in either th
...
+
+Note that if `noadjustfee` is `true`, the sender should create a transaction with RBF disabled, as the original transaction would replace the payjoin transaction.
+
+### Error format
+
+The receiver may send errors to the client, the format is the following:
+
+```json
+{
+    "httpCode": 400,
```
Kukks, line 98: The HTTP code here is pointless since this is the body of an HTTP response?

Payjoin-spec.md
```diff
@@ -70,6 +70,43 @@ There is no guarantee concerning the ordering of outputs and inputs in either th
 Note: We strongly recommend the sender to submit a PSBT in Base64 format.
 However, we also accept the PSBT or a bitcoin transaction in hex format.
 
+### Optional parameters
+
+When the payjoin sender sends the POST request to the payjoin endpoint, he can optionally specifies the following parameters:
+
+* `v=`, the version number of the payjoin protocol that the client is using. The current version is `1`.
+
+This can be used in the future so the receiver can reject a payjoin if the client is using a version which is not support via an error HTTP 400, `version-unsupported`.
```
Kukks, line 79:
```diff
- support
+ supported
```

Payjoin-spec.md
```diff
@@ -70,6 +70,43 @@ There is no guarantee concerning the ordering of outputs and inputs in either th
...
+If not specified, the receiver just assume the sender is using the same version.
```
Kukks, line 80:
```diff
- just assume 
+ will assume
```

Payjoin-spec.md
```diff
@@ -70,6 +70,43 @@ There is no guarantee concerning the ordering of outputs and inputs in either th
...
+
+If the version of the sender is above the one of the receiver, the receiver should ignore the parameter.
+
+* `feebumpindex=`, the preferred output from which to increase the fee for the added inputs. (default: `-1`)
+
+If the `feebumpindex` is out of bound or pointing on the payment ouptut to the receiver, the receiver should ignore the parameter.
```
Kukks, line 86:
```diff
- If the `feebumpindex` is out of bound or pointing on the payment ouptut to the receiver
+ If the `feebumpindex` is out of bounds or pointing to the payment ouptut meant for the receiver
```

Payjoin-spec.md
```diff
@@ -70,6 +70,43 @@ There is no guarantee concerning the ordering of outputs and inputs in either th
...
+
+* `noadjustfee=`, a boolean the client specifies if he prefers that the receiver does not bump the fee for the added input. (default: `false`)
```
Kukks, line 88:
```diff
- he prefers
+ they prefer
```

Payjoin-spec.md
```diff
@@ -70,6 +70,43 @@ There is no guarantee concerning the ordering of outputs and inputs in either th
...
+
+Note that if `noadjustfee` is `true`, the sender should create a transaction with RBF disabled, as the original transaction would replace the payjoin transaction.
```
Kukks, line 90:
```diff
- would
+ could
```

Kukks:
> 
> 
> So the response body is:
> A hex PSBT
> B base64 PSBT
> C json object with error.
> 
> Why not have it all be json?
> 
> ```js
> // success
> {
>   "success": true,
>   "format": "hex", // or "base64"
>   "type": "psbt", // or "rawtx"
>   "data": "xxxxxx(hex psbt)"
> }
> // failure
> {
>   "success": false,
>   "httpCode": 400,
>   "errorCode": "version-unsupported",
>   "message": "The version is not supported anymore"
> }
> ```

I don't agree here, the success/fail status code distinction is enough to help the implementer distinguish how to parse the response. You should assume the format of the PSBT in the response will be as the server provided it to be.

It would also be a breaking change and would be hard to coordinate between cross implementations and cross deployments.

Payjoin-spec.md
NicolasDorier: yeah

NicolasDorier:
> bump the current version to 2 (or 1.1 actually as it is not breaking) to be able to identify these parameters are available?

I don't want to overthink it, new version will know if they are compatible with older one, no need of scheme and semver. This is not something that will change a lot.

> Why not have it all be json?

Because I don't see compelling reason to break what other people already developed. This is not really important.

NicolasDorier:
For the HTTP header, I think it makes sense.

NicolasDorier:
ping @waxwing @AdamISZ @lontivero @lukechilds if that make sense and any people already implementing things.

Note that those changes don't impact what you already implemented.

I think adding as per @Kukks suggestion a HTTP header in the response of the Payjoin Receiver to signal which version he is running. It may be useful in the future.

Payjoin-spec.md
```diff
@@ -70,6 +70,42 @@ There is no guarantee concerning the ordering of outputs and inputs in either th
 Note: We strongly recommend the sender to submit a PSBT in Base64 format.
 However, we also accept the PSBT or a bitcoin transaction in hex format.
 
+### Optional parameters
+
+When the payjoin sender sends the POST request to the payjoin endpoint, he can optionally specifies the following parameters:
```
nopara73, line 75:
```suggestion
When the payjoin sender sends the POST request to the payjoin endpoint, he can optionally specify the following parameters:
```

Kukks: TIL of the github suggestion comment. Amazing

Payjoin-spec.md
```diff
@@ -70,6 +70,42 @@ There is no guarantee concerning the ordering of outputs and inputs in either th
...
+
+* `v=`, the version number of the payjoin protocol that the client is using. The current version is `1`.
```
AdamISZ, line 77: Sorry for being the dumb one around here, but I don't actually know concretely how this is implemented (probably just lack of experience with HTTP?). Currently I send a base64 PSBT in the body of the request, how do I add optional parameters like `v=`? Also in the body, I take it, but what format?

AdamISZ:
As for versioning, didn't you all tell me I was overcomplicating it worrying about versions? :) Seriously though, I think you should just do the "standard" thing even if it seems overkill: client sends version range in request, server sends response with chosen version (or error if no supported version). That just gives more flexibility for both sides in bumping versions on their own development schedule.
Protocols are not like software in this regard.

(And yes I know we don't intend to be changing this all the time, but .. it just doesn't hurt, and it could help).

Payjoin-spec.md
junderw: query parameter:

`https://btcpay.example.com/payjoin`

Would become

`https://btcpay.example.com/payjoin?v=1&noadjustfee=true`

AdamISZ: Thanks, makes sense.

Kukks:
> 
> 
> As for versioning, didn't you all tell me I was overcomplicating it worrying about versions? :) Seriously though, I think you should just do the "standard" thing even if it seems overkill: client sends version range in request, server sends response with chosen version (or error if no supported version). That just gives more flexibility for both sides in bumping versions on their own development schedule.
> Protocols are not like software in this regard.
> 
> (And yes I know we don't intend to be changing this all the time, but .. it just doesn't hurt, and it could help).

Why wouldn't the query parameter version flag not be enough? To specify a range of supported versions we could do `?v=1&v=2&v=3`
We could probably also allow specifying these via a request header instead so that the sender does not need to modify the url but maybe it's just duplicating the feature and complicating the receiver for nothing

AdamISZ:
> Why wouldn't the query parameter version flag not be enough? To specify a range of supported versions we could do `?v=1&v=2&v=3`
You wouldn't need more than two settings, min and max version supported. which of course for some clients could be the same value if they only implemented one version. I wasn't suggesting that query parameters aren't enough, just suggesting having a range client side and a choice server side.

NicolasDorier:
@AdamISZ I just want to make sure that a sender get a proper error message if there is a receiver requiring a different protocol. I prefer not doing some negotiation step, as this will be a can of worms that very few client will implement anyway. 

The goal here is to only give a way to the receiver to tell to the sender "we don't support your version".

AdamISZ:
It wouldn't be an extra negotiation step; it would be query parameters min version, max version (which can be the same of course) and receiver includes version chosen in the response.

I think a counterargument might be, that in payjoin there is actually an implicit first protocol message - the bip21 uri served (so it's not the usual protocol scenario, which starts with a client query). In which case, perhaps the server version supported range could be in the bip21uri. That way the client makes the choice, in his request, and from there both sides know the version in use.

NicolasDorier:
@AdamISZ this is a good idea.

So basically the bip21 uri, the server would include `?v=a`, and the client, if he decides to proceed would also include `v=b`? Any point for min/max version?

AdamISZ:
The purpose of the offer being a range is that software can update and not have to worry that the other side's software is not updated.

junderw:
TIL: https://stackoverflow.com/a/9547490/3074620

`https://btcpay.example.com/payjoin?myarray[]=a&myarray[]=b` will read myarray var as an array in PHP and nodeJS querystring standard libraries.

But to be specific I would suggest 3 values:

v: version (required)
vmin: minimum version allowed (default = v value) (not required)
vmax: maximum version allowed (default = v value) (not required)

AdamISZ:
I don't see a purpose in 3 params. If you only offer one version, set min and max to that same value.

junderw:
I support versions 3, 4, and 5, but I prefer 4... If the other side only supports 5, they can tell me that and we'll use it.

sender: v=4&vmin=3&vmax=5
receiver only supports 5 and above: response header version: 5

Even if we did do only two, adding the same data twice is redundant, so just make the first value required and the second value defaults to the first value if not given.

v: version client would like to use (always the highest version supported) (required)
vmin: minimum version allowed (default = v value) (not required)

etc.

AdamISZ:
Seems unnecessarily complex to me. It would be strange to a prefer a non-latest version. I mean sure it's possible to go with 3.
Stuff like TLS works like that, just client gives range, server chooses from range.
Mainly I was just arguing for having a range in the "handshake" so a heterogeneous participant set can do updates with more flexibility.

NicolasDorier:
@AdamISZ should we let the receiver announce vmin or vmax optionally? It seems unnecessary.
I agree that anyway the latest version supported should be chosen.

AdamISZ:
I don't know; if I'm writing the client (sender) code I'd like to know what the server offers. If you think it's too much space in the QR (I'll let you guys figure that out) then we could, instead, use the sequence: (1) receiver generates URI without version info and gives to sender (2) sender sends version range along with fallback PSBT (3) receiver chooses version and returns version choice with payjoin psbt. But I feel like embedding in URI is a bit better.

NicolasDorier:
I need to sleep on it.

I think it is ok, we could have a `?v=1-5` so that is only 5 chars.
If not present, the client should assume it is `v=1`. What bother me is that after somebody could say `Why not allowing a discrete versions say v=1,2,4`. (For example a receiver that disagreed with version 3)

But then the parsing rules must be parsed properly by the sender, and because this capability won't be used too much I am 90% sure senders will screw up on the parsing.

NicolasDorier:
Ok what about this we don't include the versions in the pj uri.

The sender assume the receiver is using the preferred version and sends with `v=x`
if the receiver don't support `v=x`, it returns an error with a list of supported version as a JSON array.
I think this is simpler and does not require complicated parsing logic.

NicolasDorier:
Ping @AdamISZ and @junderw check https://github.com/btcpayserver/btcpayserver-doc/pull/505/commits/f2ccd2811bce9b47f3104538158bef7983424951 I think it allows proper version negotiation without the need of crazy logic. Does this cover all cases?

AdamISZ:
ACK https://github.com/btcpayserver/btcpayserver-doc/commit/f2ccd2811bce9b47f3104538158bef7983424951

I think it is functionally very similar, probably better. If there is something worse about it I can't see it.

PR #534, opened 2020-05-18
Update FAQ-Synchronization.md

nemoBTC:
Here i propose a slightly different organization of this documentation. 

Since this is the first time i am using Github, i am not sure with the following links:
see cause 4 of [BTCPay Server keep showing that my node is always starting](#btcpay-server-keep-showing-that-my-node-is-always-starting)
In the best case this would point directly to cause 4.
And also:
[check the bitcoind log](../Troubleshooting.md#_2-1-btcpay-logs) 
seems not to point directly to 2.1 BTCPay logs


pavlenex:
Congrats on your first contribution! 🚀

PR #548, opened 2020-05-25
Fix video integration for older iOS

dennisreimann:
@britttttk mentioned on the chat, that she isn't able to play the videos on an iOS 7 device. Could reproduce it with an older iPad. Here's the fix and also a proper fallback for disabled JavaScript.


dennisreimann:
@britttttk is it fixed for you?

britttttk:
@dennisreimann yes fixed for me thanks!

Issue #512, opened 2020-05-10
CreateWallet.md


pavlenex:
Currently, a wallet can be generated in btcpay server. If `hot-wallet` or `nbx` is not ticked, this wallet is generated and the seed is read only and deleted. This means, besides `connect a wallet` which should be recommended we can have `generate a new wallet` document which is similar to HotWallet.md but explains what happens really and makes sure to clearly and simply communicates to users what they have to do (backup seed properly, no way to retrieve funds if they don't back up, etc).

This step should be discouraged in a way, but there are users who are familiar with this and would prefer (especially devs who playing around) a quick way to set up a wallet.

In our new docs, it should be here
![image](https://user-images.githubusercontent.com/36959754/81497295-d815c100-92bd-11ea-8950-e502802cba2a.png)

pavlenex:
It would be good idea while tackling this issue to do changes to wallet.md and clarify different ways to connect or generate a wallet. Add a notice that by default btcpayserver does not require a private key, though for certain use-cases (PayJoin, Liquid, Transmtuer, signing with a hot wallet) it can be used but generally is not recommended to generate a wallet within or keep the keys on the server.

PR #550, opened 2020-05-25
[WIP] Restructure ConnectWallet.md

pavlenex:
**Work in progress do not merge!**

Basically this page is a skeleton that explains some basics and helps user decide the best approach, but then provides a link towards a specific step by step setup.

Heavily relies on my idea in [1263](https://github.com/btcpayserver/btcpayserver/issues/1263#issuecomment-630893625) so should be future-proof.

Feedback welcome!

### Todo

- [x] Add `Create a wallet section` or separate it #512 
- [x] Decide how to present ColdCard.md [#1602](https://github.com/btcpayserver/btcpayserver/pull/1602)
- [x] Decide if we should rename this page to WalletSetup.md
- [x] Decide if we should have createawallet.md separately, in that case briefly mention and link to create a wallet page.
- [x] Explain why using external wallet is recommended
- [x] Add table of content links 

This is how the warnings look. Probably need to add some spacing there.

warrnings


pavlenex:
For now only need some brainstorming if this is the right direction so I can continue. @Kukks @britttttk

pavlenex:
@britttttk  Ok, this is now ready for a review. Once again the main idea is that this page introduces people to a concept of using an existing wallet and generating the one within btcpay. It provides links for each solution (hopefully the CTA's are clear enough).

After the text review, I'd like to get some feedback if we can rename this page to WalletSetup.md. I am not 100% sure what has to be changed in docs build if we rename doc (besides classic find and replace). Perhaps @dennisreimann can advise on this one.

dennisreimann:
@pavlenex Yes, a simple search 7 replace for `ConnectWallet` should work. Basically it's the links in other Markdown files and the navigation/redirects defined in the `docs/.vuepress` folder. 

sr

britttttk:
ACK for changing doc name to WalletSetup.md 

Although it seems there is quite a ways to go on this for development, I'll leave some comments for this doc. 

Only thing I will mention that I think gets left out of wallet setup is that you can't use btcpay with the bitcoin core wallet (for now afaik - not sure what the status/capability of descriptor wallets is). I think we could mention it, or direct to this [SE Q&A](https://bitcoin.stackexchange.com/questions/90135/how-do-i-export-an-xpub-from-bitcoin-core-for-use-in-btcpayserver). Possibly a Wallet FAQ instead.

docs/ConnectWallet.md
```diff
@@ -1,30 +1,61 @@
...
 - [Electrum Wallet](ElectrumWallet.md)
+- [Wasabi Wallet](WasabiWallet.md)
+
+Click on the link above you will be redirected to a step by step tutorial on how to set up particular software wallet with BTCPay Server.
+
+To spend and manage the funds received to your external software wallet, you can use an [internal BTCPay Wallet](Wallet.md) and sign a transaction with your private key or simply manage the funds in that external wallet itself.
+
+## Create a new wallet
+
+If you do not have an existing wallet, you can generate it within your BTCPay Server. If are using a [third-party host](ThirdPartyHosting.md), this option needs to be explicitly enabled by the server admin. Generating a new wallet in an environment you don't trust is discouraged.
```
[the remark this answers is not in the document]
britttttk: hmmm not sure how to word this but I think it would be better to convey the message that: if you do not know, you should consider them untrustworthy... since lots of people are thinking ThirdParty hosts are safe since they are in btcpay docs but that does not mean it's true, we do not know.

docs/ConnectWallet.md
```diff
@@ -1,30 +1,61 @@
...
+
+:::warning
+When a new wallet is generated, BTCPay Server will show you a twelve word recovery seed. After the initial display, the seed is wiped from the server, unless a [hot wallet option](HotWallet.md) is enabled.
+
+Anyone with access to your recovery seed can also access and steal all your funds. Securely back up your seed by writing it down and keep it in a safe place. Do not photograph it or store it in a digital format.
```
britttttk, line 52:
```suggestion
Anyone with access to your recovery seed can also access and steal all your funds. Securely back up your seed by writing it down and keep it in a safe place. Do not photograph it or store it in a digital format. Do not rely on your server for storing your private keys, even if you're using a hot wallet. 
```

docs/ConnectWallet.md
```diff
@@ -1,30 +1,61 @@
...
+:::
+
+Follow the [instructions here to learn how to generate a new wallet](CreateWallet.md).
+
+## Summary
+
+There are many options on how set up a wallet with your BTCPay Server. It's up to you to decide what works best for your use-case. Once you configure a wallet, you're all set. The next step, is to [start exploring the features](WhatsNext.md).
```
britttttk, line 59: >It's up to you to decide what works best for your use-case

I'm not sure about this one. Do we want to give some kind of guide or chart maybe.. Or at the very least, when creating a wallet in btcpay might be useful or chosen over the other hardware/software? Since they likely can't find that information very easily if they search something basic like: "create a new wallet in btcpay risks/benefits/tradeoffs" etc.

pavlenex: Do you have a specific example here? My idea was to provide information and let user decide based on security/ease of use. For chart do you mean like something we have for choosing deployment?

pavlenex:
Ok, replaced ConnectWallet.md with WalletSetup.md and also added a redirection so that old links with /ConnectWallet redirect to /WalletSetup so we avoid breaking changes.

Last thing we have left regarding the structure is to decide what to do with this now

sTRU

I suggest we have it like this:

Wallet Setup (WalletSetup.md)
 -- Use an existing wallet (blank)
  ---  Use an existing hardware wallet(Vault.md)
      ----AirGapped Hardware Wallet(ColdCard.md)
   --- Use an existing software wallet(blank)
      ----Electrum(Electrum.md)
       ----- ElectrumX
       ----- Electrum EPS
         - Wasabi(Wasabi.md)
 -- Create a new wallet(CreateWallet.md)
  ---  Hot Wallet(HotWallet.md)

Thoughts?

pavlenex:
So basically the new structure would look like this 
structure 1

PR #563, opened 2020-05-30
Create CreateWallet.md

### LINKED ISSUE #512: CreateWallet.md

pavlenex:
Closes #512 
- Adds a new .md file CreateWallet.md
- Adds a new path in config

Please review if warnings are clear enough and if something is unclear/missing. It's important that people are aware that they MUST back up their recovery seed which, unless `hot-wallet` is enabled is wiped from the server, hopefully I clarified that well enough.

seed


docs/CreateWallet.md
```diff
@@ -0,0 +1,23 @@
+Besides using an [existing wallet](https://docs.btcpayserver.org/WalletSetup/#use-an-existing-wallet) you can create a brand-new wallet within your BTCPay Server.
+
+If are using a [third-party host](ThirdPartyHosting.md), this option needs to be explicitly enabled by the server admin. Generating a new wallet in an environment you are not sure is trustworthy, is discouraged.
```
Zaxounette, line 3:
```suggestion
If you are using a [third-party host](ThirdPartyHosting.md), this option needs to be explicitly enabled by the server admin. Generating a new wallet in an environment you are not sure is trustworthy, is discouraged.
```

docs/CreateWallet.md
```diff
@@ -0,0 +1,23 @@
...
+Feel free to select the address format that you prefer. **Do not enable** a hot wallet, unless you've read the [Hot wallet documentation](HotWallet.md) and have the use-case for this advanced feature.
+
+![](./img/hotwallet/Setup1.png)
+
+![](./img/createwallet/createwallet.png)
+
+![](./img/createwallet/recoveryseedbackup.png)
+
+
+If you did not provide an existing seed, a seed will be generated and shown to you.The public key will automatically be imported in the store and configured to an appropriate format. If you provided an existing seed, a list of preview addresses will be shown. Validate that they are the same as the one on your wallet and Save.
```
Zaxounette, line 19:
```suggestion
If you did not provide an existing seed, a seed will be generated and shown to you. The public key will automatically be imported in the store and configured to an appropriate format. 
If you provided an existing seed, a list of preview addresses will be shown. Validate that they are the same as the one on your wallet and Save.
```

PR #577, opened 2020-06-12
chore: update docs with thub

apotdevin:


docs/FAQ/FAQ-LightningNetwork.md
```diff
@@ -334,6 +335,24 @@ export LIGHTNING_ALIAS="namehere"
...
+To install ThunderHub on your instance apply the following:
+
+```
+export BTCPAYGEN_ADDITIONAL_FRAGMENTS="\$BTCPAYGEN_ADDITIONAL_FRAGMENTS;opt-add-thunderhub"
+
+. btcpay-setup.sh -i
+```
+
+If you get the following warning message **Unable to connect to this node** apply the following:
+
```
britttttk, line 348: I think it would be better if you can mention the reason of the error or why the following commands are a solution.

docs/FAQ/FAQ-LightningNetwork.md
```diff
@@ -334,6 +335,24 @@ export LIGHTNING_ALIAS="namehere"
 . ./btcpay-setup.sh -i
 ```
 
+### How to install ThunderHub?
+To install ThunderHub on your instance apply the following:
+
+```
+export BTCPAYGEN_ADDITIONAL_FRAGMENTS="\$BTCPAYGEN_ADDITIONAL_FRAGMENTS;opt-add-thunderhub"
```
britttttk, line 342:
```suggestion
export BTCPAYGEN_ADDITIONAL_FRAGMENTS="$BTCPAYGEN_ADDITIONAL_FRAGMENTS;opt-add-thunderhub"
```

docs/FAQ/FAQ-LightningNetwork.md
apotdevin: Good point! added a more descriptive reason

apotdevin: Don't know if I should also add which domain it is but that goes a bit deeper into the docker compose setup which is maybe not needed?

dennisreimann:
Should we move these docs to the [docker docs](https://github.com/btcpayserver/btcpayserver-docker/tree/master/docs)?

britttttk:
@dennisreimann  that makes more sense to me

dennisreimann:
I think so too as we could have a separate page there. Same goes for other plugins and I'd like to feature it more prominently in the Docker Plugins section of the docs.

pavlenex:
Unless someone wants to make a proper document and explain all the functions etc, there's no reason to have a page just for a single question imo.

Issue #479, opened 2020-04-12
Document new themes darkmode / light theme in theme.md


pavlenex:
Probably would be a good idea to refresh the look of theme.md with mention of new themes and provide new images.

dennisreimann:
Yep, and maybe also a link to https://design.btcpayserver.org/components/theme-switch/

pavlenex:
#588

PR #588, opened 2020-06-17
Add server theme gif and design spec link

britttttk:
#479
- Built-in themes section for server themes
- Add reference link to standard design specs used in BTCPay


britttttk:
There are also references to these themes as predefined and preexisting. I matched those terms to be predefined and went for built-in for non-technical users with the gif. I'm open to whatever we want to call them.

PR #589, opened 2020-06-17
Setup: Windows compatibility

dennisreimann:
@britttttk mentioned some trouble with the setup script. This was related to it not being fully Windows compatible and also not supporting spaces in the file paths. Hopefully this fixes it – works for me using the Git Bash in an Windows VM.


pavlenex:
I've never had issue with the setup script on Windows, always worked via git-bash. But waiting for @britttttk to test this out.

PR #592, opened 2020-06-18
Update linkcheck and GH Actions pipeline

dennisreimann:
The linkcheck is now done locally against the current build. The pipeline also runs for pull request, so that we can ensure we do not break anything with those.
Deployment still happens only for `master` and if linkcheck succeeds. Though I also added an exclude list for external URLs that do not respond properly, this might be too restrictive. We shall see.

To test the linkcheck locally (using the production build), run the following:

```bash
npm install
./setup-deps.sh
npm run build
npm run linkcheck
```


dennisreimann:
Finally the linkcheck works. GH has a rate limit which we need to adhere to when running the crawler – so it takes up to three minutes to check all the links, but ok …

PR #603, opened 2020-06-26
Remove payjoin spec from BTCpay docs, redirect to BIP

NicolasDorier:
The specification is now part of the BIP repository https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki


Kukks:
Are you sure we should fully remove it? There are implementation specific details from our end right?

dennisreimann:
If so, we should also setup a [redirect](https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/.vuepress/redirects.js) like this:

`{ path: '/Payjoin-spec', redirect: 'https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki' },`

NicolasDorier:
@Kukks I don't think those are worth mentioning, we should add another page dedicated for that if the need arise though.

@dennisreimann how is it now?

docs/.vuepress/config.js
```diff
@@ -228,7 +228,11 @@ module.exports = {
             title: "Payjoin",
             path: "/Payjoin",
             children: [
-              ["/Payjoin-spec", "Payjoin Specification"]
+              { 
+                title: "Payjoin specification",
+                path: '/Payjoin-spec',
+                redirect: 'https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki'
+              }
```
dennisreimann, line 235:
```suggestion
              ["https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki", "Payjoin specification", { type: 'external' }]
```

dennisreimann: sorry, I meant the [separate redirects file](https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/.vuepress/redirects.js) – but let me just add it afterwards there. For the nav we need to specify it as suggested, but the case I meant was for potential links to `/Payjoin-spec` being out there …

PR #608, opened 2020-06-27
Use btcpay docs for latest release instead of master

dennisreimann:
Checkout latest release tag, so that we do not publish docs for unreleased APIs yet.

### DIFF
--- a/setup-deps.sh
+++ b/setup-deps.sh
@@ -29,6 +29,10 @@ else
 fi
 
 cd "$BTCPAYSERVER_DIR"
+
+# Checkout latest release tag, so that we do not publish docs for unreleased APIs yet.
+git checkout $(git describe --tags `git rev-list --tags --max-count=1`)
+
 if command -v jq >/dev/null 2>&1; then
   jq -rs 'reduce .[] as $item ({}; . * $item)' BTCPayServer/wwwroot/swagger/v1/*.json > "$PUBLIC_DIR/API/Greenfield/v1/swagger.json"
 fi


britttttk:
pulls released docs not master 👍

dennisreimann:
Let‘s get this merged :)

pavlenex:
Why?

dennisreimann:
Because otherwise we are publishing API docs on the website for features that aren't available in a public release yet.

pavlenex:
@dennisreimann We do releases quite often do. I really don't want to complicate the workflow, especially since we do changes in docs very often and most of them need to be visible right away.

@Kukks how do you feel about this one?

dennisreimann:
I don't see this as a big complication as we are just checking out a tag instead of master.
I've talked to @NicolasDorier about this and we both thought it'd be a good idea.

Kukks:
sorry, trigger happy, but still good

Kukks:
@dennisreimann  can we make it so that we checkout the repo twice: one for api docs and one for normal docs

Kukks:
I don't think we have any actual docs on the repo beyond the api docs though so it is fine as is.

If you have more objections @pavlenex  Ill revert the merge

dennisreimann:
@Kukks as we aren't taking any docs beside the API ones from the main repo this isn't necessary yet. But yes, we could do that or even just switch the branch again after building the API docs.

Kukks:
Ok, then this PR is completely fine for now.

pavlenex:
I don't see the point in docs having `release`. Changes in docs are immediate. Makes no sense to me to have this. But if you guys like it, then it's consensus that decided already.

Kukks:
> 
> 
> I don't see the point in docs having `release`. Changes in docs are immediate. Makes no sense to me to have this. But if you guys like it, then it's consensus that decided already.

It's not actual docs, it's the API docs. Currently it pulls from `master` which is wrong as it shows inaccurate docs for the current release

pavlenex:
Alright, haven't looked into code just read original post by Dennis. Fine then.

Issue #504, opened 2020-05-03
FAQ cleanup


pavlenex:
The FAQ are one big mess. There's plenty of duplicate, irrelevant, poorly structured and outdated info there that needs to be taken care of. Will post specifics once I dive deep into what must be cleaned. 

I made those FAQ and it takes me 10 min to  find an answer to a question I was looking for. Looking for #503 made me want to punch myself in the face.

Zaxounette:
I could try and give you a cleaned-up suggestion of the entire documentation.
Will take some time though.

All I need is the outdated info that can be deleted, so that I don't work on it for nothing :p

pavlenex:
@Zaxounette well that's the problem, someone needs to go through the outdated information and rants and remove them. I haven't looked into it yet.

What we can do is go through one by one, starting with FAQ-General.md

- [ ] Proofreading
- [ ] Removing repetitive information
- [ ] Provide concise instructions 
- [ ] Make sure all links work

FAQ-General.md in particular seems more like a blog post then a document, it's too ranty and does not provide concise information. It's also inconsistent

Zaxounette:
@pavlenex I'm willing to work on that.
Should I wait until the migration to vuepress is done ? Or can I submit PR's regarding all of this right now ?

pavlenex:
You can tackle it, @Zaxounette it does not influence the structure of what Dennis is working on.

Zaxounette:
Thanks ! =)

britttttk:
I believe Wallet FAQ is complete for this.

britttttk:
- [ ] General
- [ ] Deployment
- [ ] Synchronization
- [ ] Integrations
- [ ] Server Settings
- [ ] Stores
- [x] Wallet
- [ ] Apps
- [ ] Lightning Network
- [ ] Altcoins

pavlenex:
Great work on this one so far @britttttk!

pavlenex:
@britttttk  Lightning Network FAQ seem fine, I would leave it as is unless you noticed something I missed?

britttttk:
@pavlenex is there anything else needed for this

pavlenex:
@britttttk I don't think so, you have the honors to close this 🎉

PR #617, opened 2020-07-03
Update Stores FAQ

pavlenex:
Related #504 
Nothing much to fix here imo, I just rewrote the "invoice deleting" section. 
If someone notices more stuff to be cleaned in Stores-FAQ, let me know.

### DIFF
--- a/docs/FAQ/FAQ-Stores.md
+++ b/docs/FAQ/FAQ-Stores.md
@@ -74,7 +74,11 @@ To disable the email requirement for your store's invoices, go to Stores > Setti
 
 ## Can I delete invoices from BTCPay?
 
-Invoices in BTCPay Server can't be deleted, but can be archived. Archiving an invoice removes it from the `Invoices` page. To archive an invoice, on the invoice you would like to archive, click the `Archive` button in top right corner. An invoice can be unarchived by clicking the same button afterwards.
+Invoices in BTCPay Server can't be deleted, but can be archived. 
+To archive an invoice, select the one you would like to archive, click the `Archive` button in top right corner. 
+This action removes it from the `Invoices` page.
+
+An invoice can be restored by clicking the `Archived` button.
 
 ## How to change the exchange rate provider for invoices?
 
--- a/docs/FAQ/FAQ-Stores.md
+++ b/docs/FAQ/FAQ-Stores.md
@@ -74,7 +74,7 @@ To disable the email requirement for your store's invoices, go to Stores > Setti
 
 ## Can I delete invoices from BTCPay?
 
-No, you can't delete invoices in BTCPay. Even if the invoice is expired, invalid, paid (or any other status) you can't delete invoices because they may contain important information. For example, information from the invoice may be needed if the invoice gets paid at a later time. Try [filtering invoices](../Invoices.md#invoice-filtering) instead.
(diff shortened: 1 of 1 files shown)


docs/FAQ/FAQ-Stores.md
```diff
@@ -74,7 +74,7 @@ To disable the email requirement for your store's invoices, go to Stores > Setti
 
 ## Can I delete invoices from BTCPay?
 
-No, you can't delete invoices in BTCPay. Even if the invoice is expired, invalid, paid (or any other status) you can't delete invoices because they may contain important information. For example, information from the invoice may be needed if the invoice gets paid at a later time. Try [filtering invoices](../Invoices.md#invoice-filtering) instead.
+Invoices in BTCPay Server can't be deleted, but can be archived. Archiving an invoice removes it from the `Invoices` page. To archive an invoice, on the invoice you would like to archive, click the `Archive` button in top right corner. An invoice can be unarchived by clicking the same button afterwards.
```
Zaxounette, line 77: Tried to limit the number of "invoice" and "archive" recurrences in the § + adds spacing + rewording "restored" instead of "unarchived".

```suggestion
Invoices in BTCPay Server can't be deleted, but can be archived. 
To archive an invoice, select the one you would like to archive, click the `Archive` button in top right corner. 
This action removes it from the `Invoices` page.

An invoice can be restored by clicking the `Archived` button.
```

pavlenex:
Linkcheck failing because of `├─BROKEN─ https://raw.githubusercontent.com/dalijolijo/btcpayserver-docker-bitcore/master/btx-debian/Dockerfile (HTTP_404)` it's not related to this PR.
@dennisreimann seems like a glitch again?

britttttk:
@pavlenex it's not a bug, that broken link is fixed in the [docker repo open PR#327](https://github.com/btcpayserver/btcpayserver-docker/pull/327). Just because it's not changed in this PR doesn't mean we wont find broken links in the other repos that we pull in this check :)

britttttk:
@pavlenex I think those questions are all fine. The only thing that is confusing to me is the bullet `Store General Settings` can we make that either a new section or make the questions sub bullet points? There are examples of both of those in [Deployment FAQ](https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/FAQ/FAQ-Deployment.md)

dennisreimann:
@britttttk is right, the link ends up with a 404 indeed. It seems to have moved and we should now point to https://github.com/dalijolijo/btcpayserver-docker-bitcore/blob/master/docker-bitcored/Dockerfile

pavlenex:
> @pavlenex I think those questions are all fine. The only thing that is confusing to me is the bullet `Store General Settings` can we make that either a new section or make the questions sub bullet points? There are examples of both of those in [Deployment FAQ TOC](https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/FAQ/FAQ-Deployment.md)

We have to be careful with those, since stuff from BTCPay Server > Store > General are pointing to docs, so we must not break them if we do any sort of refactoring. I merged and will fix that broken link in another PR.

britttttk:
@pavlenex I think there is a link check in the main repo tests although I'm not certain.

pavlenex:
@britttttk There is indeed, but if we break links, those running older versions will still end up with broken links regardless.

What in particular you don't like about this part?

PR #625, opened 2020-07-08
Allow linkcheck failures in build

dennisreimann:
Removed the link check as a build/test step and added it as a periodic task. This way the link check does not interfere with the builds (which also makes them faster). It gets executed once per day, which is sufficient imho.


Kukks:
I think it would be better to have it run as part of the build but instead throw a warning not an error

Kukks:
How do we find out if the periodic link checker failed? Is it possible to have a sexy bot that creates an issue for us?

dennisreimann:
Ok, I revised it to allow failures for the linkcheck step. But hey, as everything in regards to this linkcheeck is so awesome, just when you want to test failures it works.

Issue #541, opened 2020-05-19
Add RBF to docs 


britttttk:
Maybe add to Wallet FAQ? Tooltip in the UI links to bitcoin.org RBF currently.

pavlenex:
v 1.0.5.0 introduced:

- Privacy enhancement: Randomize wallet transaction fingerprint.
- Randomize RBF support in BTCPay wallet by default for privacy

![RBF](https://user-images.githubusercontent.com/36959754/85885354-7ca78000-b7e4-11ea-843b-f8e47cc3741c.PNG)

Should be documented together with this one.

PR #573, opened 2020-06-11
Update Wallet.md

pavlenex:
- [x] Document PSBT
- [x] Document Wallet Settings (Accountkeypath)
- [x] Document BIP21 link
- [x] Document camera icon
- [x] Document coin select
- [ ] ~~Rename `Make sure no change UTXO is created` to `Don't create UTXO change`~~ done in #574 

Once complete it:
Closes #555 
Closes #477 
Closes #476
Closes #482 

This one was a bit tricky as it required a lot of tinkering and some concept changes. It's still not perfect so if anybody wants to team up, feel free to add commits or review as I work on it.

PR #634, opened 2020-07-15
Add rbf advanced settings to wallet.md

britttttk:
#541


docs/Wallet.md
```diff
@@ -82,9 +82,19 @@ Warning: Despite the fact, in this example, that you entered `1.0` in the amount
 
 ##### RBF (Replace-by-fee)
 
+Replace-by-fee (RBF) is an advanced privacy-enhancing feature that allows you to replace a previously broadcast transaction (while unconfirmed) to randomize your wallet's transaction fingerprint, or simply for replacing it with a higher fee rate to move the transaction higher in the queue of confirmation (mining) priority. This will effectively replace the original transaction as the higher fee rate will be prioritized and once confirmed, invalidating the original one (double spend).
```
Kukks, line 85: I would replace `is an advanced privacy-enhancing feature` with `is a Bitcoin protocol feature`

britttttk:
I've added RBF to the wallet doc, the remaining task from #573. 

As for the FAQ for the task of #541:

I think [this SE answer](https://bitcoin.stackexchange.com/a/54457/85016) is a bit better than the [Optech post for the RBF tooltip](https://bitcoinops.org/en/rbf-in-the-wild/) currently used in the UI, in particular because it explains the three types of RBF and the "stuck tx idea" which is more new user friendly... 

The Optech post would probably be better for the merchant issue of RBF indication.. so I am not including it but if you think it is helpful... I can add it.

pavlenex:
Nice link Brit, I find this one more helpful than the one in current UI.

PR #635, opened 2020-07-18
Add "Didn't find an answer?" link to page footer

### LINKED ISSUE #596: Add "Not satisfied? Tell us why" to docs pages
Add at the end of each doc page, add a "Not satisfied? Tell us why!" link which takes you to a new btcpayserver-doc repo github issue to write comments. @Kukks is this what you were thinking? 

I think this could be good. We don't really get much feedback on the docs other than broken links at the moment. Although the chat button is quite clear, I don't know if users will join the chat if they don't like the docs (if something was hard to find for example).

dennisreimann:
Closes #596. Adds the "Didn't find an answer?" link as proposed. 

link

This link prepopulates the issue form as displayed below. (Title and Body are prepopulated, Question label pre-assigned)

issue

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -52,6 +52,7 @@ module.exports = {
     repo: "btcpayserver/btcpayserver-doc",
     docsDir: "docs",
     editLinks: true,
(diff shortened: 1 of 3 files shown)


pavlenex:
tACK.

PR #652, opened 2020-07-24
Bonus: Add shell status script

dennisreimann:
Adds an additional "Bonus" section to the parent deployment section.

bonus


pavlenex:
Not found on this being in the structure sidebar, bonus sounds weird to me and I don't think it adds good navigation since this is the only choice. I'd add this as a FAQ question or something with themeing customization although not entirely related to that, but certainly not as page. I don't think we should clutter docs with these optional stuff.

dennisreimann:
Fair enough.

Zaxounette:
Simple though:

Instead of `Bonus`, the title in the nav bar could be `Miscellaneous`
Would sound less "weird".

docs/Bonus/StatusScript.md
```diff
@@ -0,0 +1,245 @@
...
+sudo chmod +x /usr/local/bin/btcpay-status
+
+# check if it works
+sudo btcpay-status
+
+# if so, you can also set it up as a welcome script
+sudo ln -s /usr/local/bin/btcpay-status /etc/update-motd.d/btcpay-status
+```
+
+You might need to install the dependeny `jq`, which is a prerequisite:
```
britttttk, line 25: dependency

britttttk:
I agree with @pavlenex I think this would be better suited as an FAQ and not in the nav bar.

dennisreimann:
@Kukks will integrate this directly into the docker deployment, so we don't have to have this page at all 😀

PR #654, opened 2020-07-28
Updated extended manual deployment instructions for Ubuntu 20.04

sipsorcery:
Updated extended manual deployment instructions for Ubuntu 20.04 and latest versions of application components (btc, lnd and rtl).

Adjusted steps where necessary for changes to deployment instructions for each application.

Changed the Bitcoin Core instructions to build from source rather than download. WIth this change the instructions now document a full build from source solution for BTCPay Server.

### DIFF
--- a/docs/ManualDeploymentExtended.md
+++ b/docs/ManualDeploymentExtended.md
@@ -2,6 +2,8 @@
 
 This document lists steps for manually deploying BTCPay Server and additional related components. Following these steps is likely to take a long time. A shorter and more pragmatic approach is to use a [docker based deployment](https://github.com/btcpayserver/btcpayserver-docker).
 
+The instructions also build all the application components from source which can be an advantage for certain audit and/or security scenarios.
+
 :::danger
 #### Not recommended for production use
 
@@ -10,9 +12,9 @@ Manual installation is NOT recommended for production use unless you are very co
 
 ## Installation Steps Overview
 
-The instructions in this article have been tested on Ubuntu 18.04. They should be applicable to other Linux based distributions. They are also based on all components being on the same host or virtual machine. It is possible to split the components across different hosts but these instructions don't describe that.
+The instructions in this article have been tested on Ubuntu 20.04. They should be applicable to other Linux based distributions. They are also based on all components being on the same host or virtual machine. It is possible to split the components across different hosts but these instructions don't describe that.
 
-An example hostname of `mainnet.demo.btcpayserver.org` has been used in certain places in these instructions. It needs to be replaced with the hostname you are using for your BTCPay Server.
+An example hostname of `mainnet.demo.btcpayserver.org` has been used, it needs to be replaced with the hostname you are using for your BTCPay Server.
 
 ### Security
 
@@ -21,7 +23,7 @@ If you do use these instructions to install a BTCPay Server connected to the Bit
 - [BTCPay Wallet FAQ](./FAQ/FAQ-Wallet.md)
 - [Lightning Network and BTCPay (first section)](./LightningNetwork.md)
 
-As an additional aid below is a lit of iptables rules and instructions which should include all the ports that need to be open. **NO WARRANTY**. Use at your own risk, **including risk of locking yourself out**.
+As an additional aid below is a list of iptables rules and instructions which should include all the ports that need to be open. **NO WARRANTY**. Use at your own risk, **including risk of locking yourself out**.
 
 ```bash
 ~$ vi iptables.txt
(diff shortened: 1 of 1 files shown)


docs/ManualDeploymentExtended.md
```diff
@@ -416,14 +451,13 @@ Create a bitcoin.conf file to suit your needs. An example file that is suitable
 ```
 
 ```
-mainnet=1
 server=1                              # need RPC for btcpay.
 rpcbind=127.0.0.1                     # loopback is default for 0.18.0 but no harm making sure.
 whitelist=127.0.0.1                   # for nbxplorer.
 rpcallowip=127.0.0.1/32               # loopback is default but again no harm.
-disablewallet=1                       # btcpay does not host a bitcoin wallet.
```
dennisreimann, line 424: Why remove that?

docs/ManualDeploymentExtended.md
```diff
@@ -122,17 +124,12 @@ Tor can be used by the following components to provide enhanced privacy and/or h
...
 
 ```bash
-~$ sudo apt install apt-transport-https
-~$ sudo vi /etc/apt/sources.list # (and append two lines below)
-deb https://deb.torproject.org/torproject.org bionic main
-deb-src https://deb.torproject.org/torproject.org bionic main
-~$ curl https://deb.torproject.org/torproject.org/A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89.asc | gpg --import
-~$ gpg --export A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89 | sudo apt-key add -
-~$ sudo apt update
-~$ sudo apt install tor deb.torproject.org-keyring
```
dennisreimann, line 135: Isn't this necessary anymore? Maybe it'd make sense to optionally include this for previous Ubuntu versions?

docs/ManualDeploymentExtended.md
```diff
@@ -353,17 +356,36 @@ Congratulations, all renewals succeeded. The following certs have been renewed:
 
 ```bash
 ~$ sudo nginx -v
- nginx version: nginx/1.14.0 (Ubuntu)
+nginx version: nginx/1.18.0 (Ubuntu)
 ```
 
 ```bash
 ~$ sudo sudo netstat -tlnp | grep nginx
```
dennisreimann, line 363: Double `sudo` for god mode? 😏

docs/ManualDeploymentExtended.md
```diff
@@ -960,6 +1063,10 @@ Confirm wallet password:
 17. idle      18. catch   19. robot    20. clay
 21. resemble  22. angry   23. work     24. until
 ---------------END LND CIPHER SEED-----------------
+
+!!!YOU MUST WRITE DOWN THIS SEED TO BE ABLE TO RESTORE THE WALLET!!!
```
dennisreimann, line 1067: Maybe it makes sense to highlight this even more using a ```:::warn``` block.

docs/ManualDeploymentExtended.md
NicolasDorier: Because we do not disable the wallet anymore, even on the docker deploy

docs/ManualDeploymentExtended.md
NicolasDorier: I think that's good we just put things for the latest version if this is simpler.

docs/ManualDeploymentExtended.md
sipsorcery: That's actually just a copy of the console output from running the "lncli create" command. I guess another warn block in the doc could be added but probably redundant given they'll get the same double warning when they run the command?

docs/ManualDeploymentExtended.md
sipsorcery: The recommendation on the Tor website is to use "apt install" (I can't recall the reason I didn't use that last time but most likely it was the Ubuntu 18.04 package not supporting Torv3).

sipsorcery:
> One thing I'd like to discuss: Do we want to propose building everything from source?

The only application component that previously wasn't being built from source was `Bitcoin Core`. I don't think it matters a lot whether `Bitcoin Core` is built or downloaded from a trusted source. I switched to using the build from source for completeness, since that's what all the other components were doing. 

The type of user likely to follow the steps is likely to be fairly technical and will be able to switch approaches as they desire. I've added the steps as a time saver for hunting through other documents.

NicolasDorier:
Yeah I think this doc is very important for linux people to understand how things interact with each other. As @sipsorcery said, they are savvy enough to tweak things as they want.

PR #657, opened 2020-08-02
Update each doc for standardization

Zaxounette:
This PR aims to standardize the usage of `BTCPay Server` instead of just `BTCPay`.
**Review should not be hindered by the multiple commits because it fixes this and only this.**

_Note:_ Not a single title will be changed, as to not break multiple links.

PS: Any docs not appearing as a commit in this PR has been reviewed but had no changes needed.

### DIFF
--- a/docs/Accounting.md
+++ b/docs/Accounting.md
@@ -37,6 +37,6 @@ The fields are the following:
 
 ## QuickBooks Online Connector for BTCPay Server
 
-BTCQBO is a third-party plugin that enables QuickBooks Online users to connect their BTCPay server easily. Besides bookkeeping, users can quickly generate invoices from their QuickBooks. For more information about the BTCQBO and all the features it offers, [check the plugin repository](https://github.com/JeffVandrewJr/btcqbo) or watch the video below.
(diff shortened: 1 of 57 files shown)


Zaxounette:
Ready for review.
You can hate me now !

pavlenex:
Is there really a need for this? Sometimes it's just wordiness to say btcpay server. Also conflict in  docs/LocalDevelopment.md

pavlenex:
After thinking a bit, I want to NACK this. Sometimes it's just hard to read sentences where btcpay server is mentioned a few times, btcpay or btcpay server is the same thing.

Zaxounette:
1°) 
> Sometimes it's just hard to read sentences where btcpay server is mentioned a few times,

Agreed.
Made me think of another approach to tackling this.
This PR can be closed, and in regards to your comment, i'd link to work on reducing the number of `btcpay`, `BTCPay`, `BTCPay Server` in sentences/docs, as there is way too much of them.

2°)
> btcpay or btcpay server is the same thing.

I don't agree with this.
`BTCPay Server is the product/service, as `BTCPay`, `btcpay` is a kind of slang (noun), that IMHO shouldn't be in official docs/papers or other writings.

I also recall Nicolas writing that it's BTCPay Server, not BTCPay. I'd have to find where I saw it, it was some time ago.
Hence why I standardize everywhere I can into BTCPay Server.

pavlenex:
Unfortunately me you or Nicolas don't define how people will call it, it's people who decide how they want to call sth.

PR #667, opened 2020-08-08
Hide single "ColdCard Wallet" item initially

dennisreimann:
Finally vuejs/vuepress#2408 got merged and is availabe as a release, so now we can hide the single "ColdCard Wallet" item initially. This is to prevent the assumption, that only ColdCard is supported.

Before:

before

After:

after


pavlenex:
Oh awesome! We should do this for FAQ section as well, it's cleaner.
Capture

dennisreimann:
Good point, done :)

Issue #666, opened 2020-08-08
Add infos on pj-exposed label


Zaxounette:
The following PR introduced automatic labels:
https://github.com/btcpayserver/btcpayserver/pull/1499

One label that is used is the `pj-exposed` and the information that it should provide seems unclear and introduces some questions (See [Chat](https://chat.btcpayserver.org/btcpayserver/pl/9wdcu3brxfy65c9w9hyw9xm6cw)).

That particular label could be explained specifically, but an idea could be to add every single automatic label to the docs, and expand on the meaning of the individual labels, as to clear things for the user.

Maybe afterwards, add an interrogation marker in the wallet page in BTCPayServer directing to that docs page ?

Zaxounette:
New comment related to this issue: https://github.com/btcpayserver/btcpayserver-doc/pull/670#issuecomment-674354579

pavlenex:
@britttttk @Zaxounette is this closed by #670? It seems Brit's PR resolves is on docs end.
Would you like to add a tooltip in BTCPay's UI or?

Zaxounette:
#670 was only part of the issue.
It resolves the issue of not having a table with the transaction labels in the docs. Now we do ! 

But there's still the `pj-exposed` label that is unclear.
Still need @Kukks 's input on this to clarify what information he wanted to send a user when labeling `pj-exposed`.

britttttk:
@pavlenex yes I will try to describe the labels a bit better from looking at the bip, etc. before we add the tooltip as docs are just matching the ui copy at the moment.

Kukks:
Hey, sorry overlooked this discussion.

`pj-exposed` is marked on utxos/transactions that were exposed to customers when they were attempting a payjoin transaction. 

* I have 1utxo in my store wallet 
* I enable payjoin 
* a customer generates an invoice to pay and has a payjoin compatible wallet
* customer initiates a payjoin transaction.
* the utxo in the store is used to create the payjoin and is exposed to the customer.
* The customer knows that that utxo is under ownership of the store. 
* It is irrelevant is the payjoin is completed or abandoned, the utxo was exposed to at least one external user, I would discourage usage of that utxo to anything sensitive (or consolidating it with other utxos that were not exposed) if you are motivated to protect your privacy.
* The coin selection wallet feature is really handy to deal with these exposed coins

Zaxounette:
Nice, thanks a lot @Kukks ! :)

britttttk:
[...]

PR #670, opened 2020-08-15
Add transaction labels table to wallet doc

britttttk:
Start #666

Just copied the ui descriptions for now. I agree @Zaxounette we should add a tooltip to the payjoin doc in the ui. I couldn't find it.


Zaxounette:
Thanks britttttk, that's neat and tidy ! 

The next thing we could discuss is how to make the sentence `Utxo was exposed through an invoice payjoin proposal` clear for users, me being the first.

I have no idea what is meant by the UTXO being `exposed`. 
Exposed in what sense ? On the blockchain, leaking linkage between UTXO's ? 
Is it just a signal for a change address not to be used again in a payjoin because it `will` link to previously completed payjoin ?
Something else ?

I'm gonna ping @Kukks so that we can work with his input on the meaning of the term he had in mind as well.

Issue #671, opened 2020-08-15
Breaking Contributing.md into sub-sections in order to define processes more clearly


pavlenex:
Ubuntu has a very nice documentation of processes for new contributors. I think the way we now have them is a bit too generic, perhaps we can break it down into specific sections and provide clear processes on how to contribute to a certain part of btcpay dev / doc / design/ seo / etc

I think for new contributors to get involved they need clear instructions and well-documented process, I think we can break that down in docs

http://ubuntu-manual.org/getinvolved

I would probably remove  Hosting Providers or keep such as on a generic page.
Contributing.md should probably be a page to help encourage new contributors where as all the sub-pages should define simply the process and tools for each "team".

If you're aware of similar pages in foss, let's share them to brainstorm a solution that would fit our community.

Zaxounette:
I'm willing to draft something up tomorrow in the morning,

britttttk:
I think WriteStack should be changed, to WriteSoftware... it does not make sense to me how it is now. I also think that Write in the document title is unnecessary and is confusing, WriteDirectory and WriteStack are examples of this...

docs/Contribute/ContributeWrite/WriteDocs.md  words are repeating here, seems redundant

pavlenex:
I tend to agree with @britttttk , I mentioned to Zax that writers are not likely to do a software change, even if it's a typo.

Which title are you referring to, the filenames?

britttttk:
Actually I'm thinking about it now, I understand the reason for the Write in the titles.. this makes sense to me. We would also then always know that it's a contribution doc by the Write in the title.

Zaxounette:
> I think WriteStack should be changed, to WriteSoftware... it does not make sense to me how it is now.

Sure I can do that if it's confusing.
But keep in mind, it writing **for** the software, not writing **the** software.

Zaxounette:
> > I think WriteStack should be changed, to WriteSoftware... it does not make sense to me how it is now.
> 
> Sure I can do that if it's confusing.
> But keep in mind, it writing **for** the software, not writing **the** software, that'll be covered in ContributeDev.

FYI, done in #684

Zaxounette:
@britttttk , following your comment and link to your old branch ([here](https://github.com/btcpayserver/btcpayserver-doc/pull/684#issuecomment-683177525))
I'd like to get it into the docs for the ContributeCode section (yes, renamed it, not ContributeDev anymore) even if ther's still some TODO's at the bottom.

[...]

PR #686, opened 2020-08-28
Contribute.md refactoring - Create WriteDocs.md

Zaxounette:
Part of #671 

* Creates WriteDocs.md inside Write folder

_config.js will be updated once all Write docs are created._

### DIFF
--- a/docs/Contribute/ContributeWrite/WriteDocs.md
+++ b/docs/Contribute/ContributeWrite/WriteDocs.md
@@ -5,10 +5,4 @@ Helping us keeping the present [Documentation](https://github.com/btcpayserver/b
 Beginners can view the following video on how to contribute to the BTCPay Server documentation:
 [![Contributing to Documentation](https://img.youtube.com/vi/bSDROcdSSWw/mqdefault.jpg)](https://www.youtube.com/watch?v=bSDROcdSSWw "How to contribute to BTCPay Server documentation")
 
-Contributing to documentation follows the same process as contributing for the Software Stack.
-
-**Step 1**
-
-Fork/Clone the main documentation repository ([BTCPay Server Documentation](https://github.com/btcpayserver/btcpayserver-doc)) using Github Desktop and publish it.
-
-**Then** follow **steps 2 through 5** of the [Software Stack](./WriteStack.md) contribution page that explains creating branches and modifying files with Github.
+The main documentation repository to fork/clone is the following: [BTCPay Server Documentation](https://github.com/btcpayserver/btcpayserver-doc)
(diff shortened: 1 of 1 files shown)


docs/Contribute/ContributeWrite/WriteDocs.md
```diff
@@ -0,0 +1,14 @@
...
+Beginners can view the following video on how to contribute to the BTCPay Server documentation:
+[![Contributing to Documentation](https://img.youtube.com/vi/bSDROcdSSWw/mqdefault.jpg)](https://www.youtube.com/watch?v=bSDROcdSSWw "How to contribute to BTCPay Server documentation")
+
+Contributing to documentation follows the same process as contributing for the Software Stack.
+
+**Step 1**
+
+Fork/Clone the main documentation repository ([BTCPay Server Documentation](https://github.com/btcpayserver/btcpayserver-doc)) using Github Desktop and publish it.
+
+**Then** follow **steps 2 through 5** of the [Software Stack](./WriteStack.md) contribution page that explains creating branches and modifying files with Guthub Desktop.
```
pavlenex, line 14: I'm not a fan of pointing people to different page for the steps. Should we eliminate the written steps entirely since the video seems to be enough?

Zaxounette: I feel like the more ways we show (written + video is better than just written or just video ?) the better ?

Would it be suitable to just copy the writewtack.md steps into here (and the other docs) as to eliminate the link ?

Zaxounette:
```suggestion
**Then** follow **steps 2 through 5** of the [Software Stack](./WriteStack.md) contribution page that explains creating branches and modifying files with Guthub.
```

Zaxounette:
```suggestion
**Then** follow **steps 2 through 5** of the [Software Stack](./WriteStack.md) contribution page that explains creating branches and modifying files with Github.
```

Zaxounette:
@pavlenex I feel like the more ways we show (written + video is better than just written or just video ?) the better ?

Would it be suitable to just copy the writewtack.md steps into here (and the other docs) as to eliminate the link ?

pavlenex:
I'm not really a fan of copying steps as each is in a way unique to specific contribution, and can cause frustration if in the future we have multilingual docs, imagine being a translator having to translate same set of string for 3 times.

Zaxounette:
Done ! 
Kept the last sentence and rephrased it, just so that a clickable link to the relevant repo be available.

PR #687, opened 2020-08-28
Contribute.md Refactoring - Create WriteDirectory.md

Zaxounette:
Part of #671 

* Creates WriteDirectory.md in Write folder

config.js will be updated once all Write docs are created.


pavlenex (commented):
Does directory really fall under writing? I don't feel like it does. It's a different monster?

Zaxounette:
> Does directory really fall under writing? I don't feel like it does. It's a different monster?

Okay I see what you mean and I concur.
I'll dynamite this and keep it for another section, potentially `ContributeMisc` (As in Miscellaneous) for the "little stuff" that wouldn't get more than a few lines like `reporting issues`, being a `3rd party host` or `spreading the word`.

PR #693, opened 2020-09-04
Contribute devCode doc

britttttk:
Part of #671

PR #694, opened 2020-09-04
Move ContributeDevCode.md from Contribute folder to ContributeDev folder

Zaxounette:
Follow-up on #693 to get the new doc in the proper structure.

Will update config.js once ContributeDevCode and ContributeDevTest are in their final versions.


Zaxounette:
@pavlenex

PR #703, opened 2020-09-12
Hot wallet warning

britttttk:
Add danger warning to hotwallet.md for sharing server credentials with others.


pavlenex (commented):
Can you put the warrning below the bullet points, it's a bit weird that the warning goes after the `:`

pavlenex (approved):

PR #704, opened 2020-09-12
Change absolute URLs to relative URLs

l33d4n:


britttttk:
hey @l33d4n thanks for the link fixes! There have been a few updates to the link processes recently. It looks like from your PR we should not be using the docs.btcpayserver/guide style links anywhere, we should be using the github relative links everywhere right?

We might want to make a note of this in the contribute writing doc if that is the standard. I'm not sure this is obvious for new contributors. What do you think? @Zaxounette @dennisreimann

l33d4n:
@britttttk The absolute URLs do not work properly locally. They display the pages from docs.btcpayserver.org instead of the local pages.

dennisreimann:
As @l33d4n said it should be done using the relative paths where possible to make it work in all cases. [Vuepress transforms those links](https://vuepress.vuejs.org/guide/markdown.html#links) for the website and they work on GitHub too when browsing the repo.

pavlenex (approved):

PR #707, opened 2020-09-13
Show social icons on mobile below the fold

### LINKED ISSUE #692: Show social icons on mobile below the fold
A minor nit I noticed while browsing docs is that our mobile menu shows social icons + website and chat links above the fold which isn't optimal UX imo. I suggest we show them below the navigation links and perhaps we can show social icons in a single row?

Optionally maybe we can show them in the header, but I'm unsure how would that look.

Screenshot 2020-09-01 at 19 24 39

l33d4n:
Close #692 

| Before  | After |
| ------------- | ------------- |
image  | image

| Dark Before  | Dark After |
| ------------- | ------------- |
image  | image


pavlenex (approved):

pavlenex:
Awesome, thanks for the contribution spree 🚀

PR #725, opened 2020-09-21
Add Invoice translation § to btcpayvsothers.md

Zaxounette:
Kukks mentioned in a recent tweet the invoice translator that is built in BTCPay Server.
It made me realize I couldn't recall if I saw mention of it in the docs, and searching in them resulted in nothing.

* Adds a § mentioning invoice translating in `invoice.md`

Unsure if I should add a screenshot. Doesn't seem absolutely necessary as usage is straightforward.

### DIFF
--- a/docs/BTCPayVsOthers.md
+++ b/docs/BTCPayVsOthers.md
@@ -61,10 +61,15 @@ First rule of Bitcoin is always keep your private keys *private*. Using a secure
 Secondly, there is another area of security to consider on the applications layer where you have two main options:
 
 * **Option 1**: Most payment processors (including BTCPay Server) use the [BIP 21][1] standard.
+
 * **Option 2**: Others use variations of the [BIP 70][2] standard.
-     * **Note**: [BIP 70 has recently been deprecated in Bitcoin Core][3].
-     * Many wallets do not allow payments to BIP 70 invoice urls.
-     * Need help converting obsolete BIP 70 invoice urls to a BIP 21 invoice url? No problem, we can help with that too.<br>Here's our [BIP 70 invoice url converter][4].
+**Note**: [BIP 70 has recently been deprecated in Bitcoin Core][3].
+Some payment processors still use this depreciated standard, but some wallets will not be compatible. To be able to pay to a BIP 70 invoice, you need to translate it.
+In order to do so, every BTCPay Server instance also hosts a translation tool, that you can access by adding `/translate` to your BTCPay Server URL.
+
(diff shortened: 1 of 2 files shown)


pavlenex (commented):
To me invoice.md is page that is useful for invoice management. 99% of people needing invoice info won't care about this use-case imo, I'd add it to FAQ or something. Not really sure where to put it.

Zaxounette:
ACK on the placement.

Need to crunch on where to put it then.

Can't go in app.md for the same reason as you mentioned for invoice.md
There's no FAQ page that it could relate to either.

What about creating a Misc FAQ page for stuff that relates to nothing else ?

britttttk:
It exists in docs [here](https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/BTCPayVsOthers.md#security) to use the mainnet one. Personally I think this link should be listed in the btcpay interface in external services but maybe that's too much clutter. 

I just don't think anyone will ever search for something if they have never seen it in their btcpay. But I guess FAQ would be good. Maybe we should make something in the docs that says all things your btcpay routes to by "default". Is there anything else like this?

pavlenex:
In that case @britttttk I would just modify this part and explain that each btcpayserver instance has a `/translate`. This is a minor feature that not many people would need so no need to display it prominently in the docs imo.

Zaxounette:
> It exists in docs [here](https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/BTCPayVsOthers.md#security) to use the mainnet one. 

Wow.
You sure weren't a detective in another life ? That one was buried ! 

So the translator is mentioned in the docs, but not in the stack UI anywhere.
I'll commit what you suggested @pavlenex , adding the current PR's text into the paragraphe @britttttk linked in BTCPay vs others.

Also, adding the translator to services (internal since it's already packaged like RTL for example) would indeed be a good idea, since it's not a well "marketed" feature. I'll just create an issue for that in the BTCPay Server repo.

pavlenex:
> Also, adding the translator to services (internal since it's already packaged like RTL for example) would indeed be a good idea, since it's not a well "marketed" feature. I'll just create an issue for that in the BTCPay Server repo.

It's not really a service though. So I doubt that one will get a consensus to get added there, let's wait for a plugin system that may happen in the future maybe it'll make more sense there, as indeed the way it is now, nobody knows about it nor uses it.

Zaxounette:
Okay so i'll push changes here for what you suggested and create the issue in appropriate repo.

Will be done by tonight.

pavlenex (approved):

PR #729, opened 2020-09-24
Fix latest tag lookup for btcpayserver repo

dennisreimann:
Refines the lookup for the latest release tag for btcpayserver , so that only API docs for released versions get published.
The previous version incorporated all tags, assuming there are only release tags, which isn't the case.

This change ensures the tag name starts with `v[0-9].`

### DIFF
--- a/setup-deps.sh
+++ b/setup-deps.sh
@@ -40,7 +40,7 @@ for file in "$DOCS_DIR"/BTCPayServer/*.md; do
 done
 
 # Checkout latest release tag, so that we do not publish docs for unreleased APIs yet.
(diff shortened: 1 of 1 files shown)


dennisreimann:
The "Setup dependencies" build step now reeports:

```
Run ./setup-deps.sh
Cloning into '/home/runner/work/btcpayserver-doc/btcpayserver-doc/deps/btcpayserver'...
Note: switching to 'v1.0.5.5'.
```

Looking good.

Kukks (approved):
spiritual approval from me as im not fluent in these commands

PR #731, opened 2020-09-24
Fix missing social icons

### LINKED ISSUE #730: 2 Social icons missing on destkop but appear in mobile menu
2 social icons (for our Mattermost and Website) are missing from the desktop version of our docs, where as they appear on mobile.

Screenshot 2020-09-24 at 19 33 57

Screenshot 2020-09-24 at 19 38 10

Probably caused by #707 Unfortunately at that time I didn't run the dev env to test this as my GPU died.

Ping @l33d4n is there a particularity reason why you prefer to hide these icons on desktop?

l33d4n:
Closes #730 

image


britttttk:
Nice! Way better to have MM and the website link easier to find on desktop site.

dennisreimann:
Updated it a bit, because I think we can just remove the rules introduced in 14ac218 instead of overriding them again.

Also improved the alignment of the navbar while wee are at it :)

dennisreimann (approved):

l33d4n:
I have no idea how I missed it, thanks a lot! :)

PR #733, opened 2020-09-26
Maintenance

dennisreimann:
Updated the dependencies and design files. Just want another pair of eyes on this to verify nothings broken :)

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -228,13 +228,13 @@ module.exports = {
           ["/Apps", "Apps"],
           ["/Wallet", "Wallet"],
           ["/Invoices", "Invoices"],
-		  {
-		    title: "Pull Payments",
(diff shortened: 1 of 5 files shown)


pavlenex (commented):
Tested a bit and the only thing that does not work on my end is not being able to see items under development sidebar menu and scroll past it.
Screenshot 2020-09-26 at 21 18 32

dennisreimann:
@pavlenex That works for me. Did you run the `./setup-deps.sh` script? The first subitem in development is "Architecture", which is from the external BTCPay Server repo – I suspect it not being present being the culprit.

pavlenex (approved):

pavlenex:
Ugh yeah, I changed my dev env and completely forgot about that one (though it did work on a different PR). Since you don't have these problems, it's good to go then.

Issue #726, opened 2020-09-22
Add update alert option to docs


britttttk:
Possibly add this to [How to update FAQ](https://docs.btcpayserver.org/FAQ/FAQ-ServerSettings/#how-to-update-btcpay-server) or add new FAQ how to be notified for updates.

![image](https://user-images.githubusercontent.com/39231115/93847143-57030900-fc63-11ea-850e-edfa21d712d2.png)

pavlenex:
I think it's not just the box that needs to be enabled, @rockstardev do we need to set up an env variable for this to work or should it work if the checkbox is selected out of the box?

britttttk:
I think it should work OOTB and it checks once per day, afaik 24 hours after the time it's enabled.

PR #737, opened 2020-10-01
Mention update alert policy in version FAQ

britttttk:
#726


britttttk:
Details about this feature are [here](https://github.com/btcpayserver/btcpayserver/pull/1788). 

I don't think we need to document specifics of it (checks only once per day... etc). I didn't use the entire policy name, it's quite long. `Enable it in Server Settings > Policies > Check releases on GitHub and alert when new BTCPayServer version is available`

Zaxounette (approved):

britttttk:
I did not see the update alert by doing this.

Zaxounette:
> I did not see the update alert by doing this.

I concur.
I had the update checkbox clicked before the last release, and waited 24+ hours to see if I would be an alert. Didn't get anything either.

rockstardev:
Nicolas asked for variable in config settings so you need to populate `BTCPAY_UPDATEURL` to be `https://api.github.com/repos/btcpayserver/btcpayserver/releases/latest` for this policy to be usable.

See here: https://github.com/btcpayserver/btcpayserver/pull/1788#discussion_r464217229

rockstardev:
Seems that @NicolasDorier already merged it so if you pull latest master it should work from now on.

Do mention this variable in docs for paranoid crowd, in case they want to disable it so that server doesn't ping github. The way it's setup after merge is that server will start autochecking for new version on all new installations because BTCPAY_UPDATEURL var is present.

Zaxounette:
Okay so If I understand correctly, the doc should state that starting in BTCPayServer v1.0.5.6 the server owner will be automatically notified, but for previous versions (up to when the checkbox is included) have to SSH and edit BTCPAY_UPDATEURL ?

rockstardev:
No, for existing server owners - they first need to pull the `btcpayserver-docker` repo and start with `BTCPAY_UPDATEURL`. So likely they'll receive notification only after v1.0.5.7... that is if they turn on checkbox because for them `CheckForNewVersions` checkbox will be false.

It'll be earlier for those that manually pull master - they'll receive notification for v1.0.5.7 if they set `CheckForNewVersions` to true.

Zaxounette:
Okay so in any case it'll be an automatic feature once 1.0.5.7 is released ?

rockstardev:
😄 unfortunately - no, not universally. For existing installations, they'll have `CheckForNewVersions` set to false, so those users will need to opt-in by checking the box.

I've created a PR that will hide the `CheckForNewVersions` checkbox if `BTCPAY_UPDATEURL` env var is not present: https://github.com/btcpayserver/btcpayserver/pull/1939

Zaxounette (approved):

pavlenex (approved):

PR #751, opened 2020-10-18
Doc configurator

### LINKED ISSUE #446: Document Web Deployment via Configurator
* Rename Web Deployment to "Luna Node Web Deployment"
* Document using install.btcpayserver.org
* Document risks of using hosted
* Document running it locally instead

britttttk:
Close #446

- install.btcpayserver.org gives me an errors page and doesn't work when I tried to use it so I do not mention it.
- I don't make a reference to this in the ThirdParty doc because it remembers the previous deployment's ssh credentials.. I wish it didnt.. 
- I can include it [here](https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/WhatsNext.md#connecting-additional-btcpay-services) if we want.
- If this moves to a plugin that's fine but this is an old issue so here is the doc also it works really well for any btcpay consultants :) 
- I can add it to deployments doc but it's not really for _first deployments_ so that might be confusing.. any ideas?


dennisreimann:
Awesome, good job! 

The only thing I'd consider is moving these docs to the [configurator repo](https://github.com/btcpayserver/btcpayserver-configurator) (as it's README) and integrate it. This is how we've done it with the docker and transmuter repos as well and I think it'd work well as the configurator README, which currently isn't available. This way people coming to the repo would benefit without having to visit the docs too.

Kukks:
> Close #446
> 
> * install.btcpayserver.org gives me an errors page and doesn't work when I tried to use it so I do not mention it.
> * I don't make a reference to this in the ThirdParty doc because it remembers the previous deployment's ssh credentials.. I wish it didnt..
> * I can include it [here](https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/WhatsNext.md#connecting-additional-btcpay-services) if we want.
> * If this moves to a plugin that's fine but this is an old issue so here is the doc also it works really well for any btcpay consultants :)
> * I can add it to deployments doc but it's not really for _first deployments_ so that might be confusing.. any ideas?

> * install.btcpayserver.org gives me an errors page and doesn't work when I tried to use it so I do not mention it.
What's the error? I'd rather we point people to use that instance over third parties that deployed it.

> * I don't make a reference to this in the ThirdParty doc because it remembers the previous deployment's ssh credentials.. I wish it didnt..
Only the browser remembers it with the form autofill stuff.

> * If this moves to a plugin that's fine but this is an old issue so here is the doc also it works really well for any btcpay consultants :)
It will be both. When in plugin mode, it will additionally allow server admins to edit the current btcpay server setup. 

> * I can add it to deployments doc but it's not really for _first deployments_ so that might be confusing.. any ideas?
I think this tool is especially for first time deployments. The whole point is to give an easy UI installer experience not tied to a specific VPN (LunaNode ) service.

britttttk:
>What's the error? I'd rather we point people to use that instance over third parties that deployed it.

I agree, pointing people to install.btcpay would be good. I've opened an issue with the exception info in the configurator repo. 

@dennisreimann @Kukks thanks for the info and suggestions! I'll move this and update it from your comments and link it back to these docs.

pavlenex:
I wouldn't point people to that URL, as we still use it to test stuff. It can create a big attack vector and we haven't considered security implications of it. Related discussion around this on [Twitter](https://twitter.com/MrKukks/status/1216033137114959872?s=20).

Kukks:
I  disagree, I would rather link to that one managed by us, at least we
know that it is not some phishing site. Credentials are only stored briefly
and aren't logged afair

dennisreimann:
@britttttk Let me know in case you need help with moving it – I could also take care of it after this got merged.

pavlenex:
@dennisreimann since this is now merged in the configurator readme, perhaps you'd want to add it to docs?

dennisreimann:
Will provide a PR for that :)

PR #753, opened 2020-10-19
Update dependencies

dennisreimann:
The regular dependency upgrade that could use another pair of eyes verifying it didn't break something.


pavlenex (approved):
tACK. No breaks occurred during my brief test.

PR #763, opened 2020-10-29
Update Support.md

Eskyee:
Available for paid support


britttttk (changes requested):
@Eskyee can you add yours after `Andrew Camilleri(Kukks)` to keep the list in alphabetical order.

Issue #773, opened 2020-11-22
Documentation updates for store settings changes


britttttk:
Find and update documentation copy/image references to the Derivation Scheme section in the store settings page. The appearance of it/button names have changed for the next major release:

![image](https://user-images.githubusercontent.com/39231115/99898522-19b8f880-2c5f-11eb-98c8-562a37ba4684.png)

Previous:
![image](https://user-images.githubusercontent.com/39231115/99898627-30138400-2c60-11eb-890e-e59b91cc395e.png)

Notes:
"Derivation scheme" term is still used on the setup page. 
The page has also been re-arranged but I don't think it causes docs conflicts.

Zaxounette:
While looking at where to include such images, I mapped this: 

Pages relating to this issue that have no images to replace:
* https://docs.btcpayserver.org/WalletSetup/#use-an-existing-wallet

Pages relating to this issue that have images that can be replaced:
* https://docs.btcpayserver.org/ElectrumWallet/
* https://docs.btcpayserver.org/WasabiWallet/

pavlenex:
This should be done ASAP, as it's an important change in the UI that may confuse newcomers.

britttttk:
> https://docs.btcpayserver.org/WalletSetup/#use-an-existing-wallet

I didn't find ay on this page but I did find some updates in CreateWallet

britttttk:
This PR updates:
- @Zaxounette page mentions above
- Updates a derivation scheme page image used in the Wasabi and Electrum docs with  Wasabi specific text on it.
- Setup Wallet: copy updates to: Stores >  Settings > Wallet > Setup
- Search everywhere in docs for "Derivation" to find the wallet connection process

If anyone finds more places please let me know I think this should be complete.

PR #778, opened 2020-11-25
connect xpub ui updates in docs

britttttk:
part of #773


pavlenex (approved):

pavlenex:
This is good start, but we have to dive deep into docs and find mentions of `derivation scheme` and `modify` since tons of docs refer to that, though since we'll be doing lots of changes in the wallet setup in the next release, this is not super-urgent, so I'll merge this one for now.

PR #783, opened 2020-11-28
FAQ - Store & PoS redirects

### LINKED ISSUE #760: Document PoS custom invoice redirect
Added in [#1924](https://github.com/btcpayserver/btcpayserver/pull/1924) there are now custom redirects in PoS

image

- Example use case: If you want to have your PoS donation page redirect to a social media or thank you page, you can do this by adding the redirect url in your PoS App. 

I suggest adding this to the Apps FAQ.

britttttk:
closes #760 
- Add store redirect info to Store FAQ
- Add PoS redirect info to App FAQ


britttttk:
@Kukks @mariodian can you take a look at the options for PoS redirects I have documented, [here](https://github.com/btcpayserver/btcpayserver-doc/pull/783/files#diff-c71c1e749d8ac6e1a3f0ac416de0baf161b1a47c44f2e8a897bccdc1ab040d56R108) it seems like there is no point to have the `Use Store Settings` option, since when it is enabled, it would be the same as `Yes - Without providing Redirect URL`.

Was there a place to provide the redirect URL at the store level previously that was removed?

Kukks:
> @Kukks @mariodian can you take a look at the options for PoS redirects I have documented, [here](https://github.com/btcpayserver/btcpayserver-doc/pull/783/files#diff-c71c1e749d8ac6e1a3f0ac416de0baf161b1a47c44f2e8a897bccdc1ab040d56R108) it seems like there is no point to have the `Use Store Settings` option, since when it is enabled, it would be the same as `Yes - Without providing Redirect URL`.
> 
> Was there a place to provide the redirect URL at the store level previously that was removed?

You can provide a redirect URL on an invoice level, meaning when it is created. This is where `Use Store Settings` acts as a default unless `redirectAutomatically` was provided to the invoice creation. Apps with no redirectURL set can also be overridden when their associated invoice is created (through the API documented on the `Edit App (bottom section)` ). Payment buttons can also specify them.

There was never a place to entire a store-level redirect URL but we could add it in.

britttttk:
@Kukks, thanks for clarifying! I think I've made it more clear now wrt your comments.

pavlenex (approved):

PR #785, opened 2020-12-01
Setup wallet from QR Code xpub

### LINKED ISSUE #776: Document wallet import via QR code
Not of a major urgency, since we'll probably have [changes in the wallet setup](https://github.com/btcpayserver/btcpayserver/issues/1703) in the next release, but would be good to keep a note on the ability to import a wallet via a simple QR code scan. From what I've read BlueWallet and Cobo Vault have the ability to export  xpub+fingerprint via QR code which apparently should be added in the wallet.

Haven't had much luck testing with BlueWallet so far, so keeping this open if someone wants to give it a try.
QR Code

britttttk:
Closes #776

### DIFF
--- a/docs/FAQ/FAQ-Wallet.md
+++ b/docs/FAQ/FAQ-Wallet.md
@@ -12,6 +12,7 @@ This document contains frequently asked questions related to BTCPay Server's [in
 * [How to add custom labels and comments to transactions?](#how-to-add-custom-labels-and-comments-to-transactions)
 * [I don't see Lightning network payments in BTCPay wallet?](#i-dont-see-lightning-network-payments-in-btcpay-wallet)
 * [Is there a mobile app for BTCPay Server wallet?](#is-there-a-mobile-app-for-btcpay-server-wallet)
+* [Can I connect my mobile wallet to BTCPay?](#can-i-connect-my-mobile-wallet-to-btcpay)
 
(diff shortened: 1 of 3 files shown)


britttttk:
I haven't tried Cobo wallet but I can add it to the doc. I just wasn't sure which had the xpub export, their mobile or their hw wallet.

pavlenex:
Are you sure this is the right place to document this?

What this does is basically imports your xpub and sets up wallet fingerprinting and account key path automatically, so I'd rather see this added somehow in `Use an existing software wallet`?

britttttk:
>Use an existing software wallet

I can do that. I just think it adds clutter to the wallet setup page with this info about Import From .. QR Code, fingerprint, etc. Which the other software wallets have their own doc.

pavlenex:
@britttttk Yes, we'd need a separate page most likely. I think in the next release, we will try to distinguish generate wallet page and import wallet happening on the same page, so perhaps that can be a new page, where user can import a wallet instead generating one or scan via QR, then we can perhaps document this properly?

dennisreimann:
I think we should move some of those paragraphs right onto the new import page. This way people find the instructions on the particular import method right away, whichout consulting the docs. I'll tinker with that.

britttttk:
Closing this, to be addressed in #837

PR #806, opened 2020-12-22
Ypub-Zpub to Xpub conversion

### LINKED ISSUE #805: Use extended public key instead of xpub because we support ypub and zpub
We mentioned xpub so many times in documentation. In reality it's not only xpub that's supported. We also support ypub and zpub. It would be good to distinquish those somehow and make clear which formats are supported and that we also [convert ypub and zpub to xpub.](https://github.com/btcpayserver/btcpayserver/discussions/2163) https://support.samourai.io/article/49-xpub-s-ypub-s-zpub-s

Zaxounette:
Closes #805 

* af12968, 788b59e and bd185c6 simply replaces `Xpub` by `extended public key` in different files where feasible. 
Excludes image and link titles or adresses.
* c7a1551 Adds a `tip` in `WalletSetup.md` briefly explaining that BTCPay converts Ypub and Zpub to Xpub without incidence feature-wise.
* 741e1dc Adds another `tip` in `FAQ-Wallet.md`
* 182a1ab Replaces `Xpub` by `extended public key` where relevant

I kept this simple, as to not overwhelm users that could be uneducated to the specifics of these formats. 

It may be too simple.
Thoughts ?

### DIFF
--- a/docs/FAQ/FAQ-Wallet.md
+++ b/docs/FAQ/FAQ-Wallet.md
@@ -91,7 +91,7 @@ Using different derivation schemes with your extended public key, you can also c
 |Multi-sig P2SH |	2-of-xpub1...-xpub2...-[legacy] |
 
 :::tip
-On top of the Xpub extended public key formats shown above, BTCPay Server supports Ypub and Zpub formats. Please note that these will be converted to Xpub once the wallet setup is completed. This has no effect on how you receive or send funds.
+On top of the xPub extended public key formats shown above, BTCPay Server supports yPub and zPub formats. Please note that these will be converted to xPub once the wallet setup is completed. This has no effect on how you receive or send funds.
 :::
 
 ## What is a Replace-By-Fee (RBF) transaction?
--- a/docs/FAQ/FAQ-Wallet.md
+++ b/docs/FAQ/FAQ-Wallet.md
@@ -33,7 +33,7 @@ This  means that you're using a hardware wallet without leaking information to t
 
 ## Do I have to use BTCPay Server wallet?
 
-By default BTCPay Server only requires an extended public key. To receive payments to your BTCPay store, you need to provide an extended public key (xpub) which you can generate in an external (existing) wallet. You do not have to use the built in wallet at all, you can manage funds in your [existing wallet](../WalletSetup.md#use-an-existing-wallet).
+By default BTCPay Server only requires an extended public key. To receive payments to your BTCPay store, you need to provide an extended public key which you can generate in an external (existing) wallet. You do not have to use the built in wallet at all, you can manage funds in your [existing wallet](../WalletSetup.md#use-an-existing-wallet).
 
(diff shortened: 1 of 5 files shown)


pavlenex:
Would be a good idea to document it here https://docs.btcpayserver.org/FAQ/FAQ-Wallet/#what-is-a-derivation-scheme as well.

Zaxounette:
> Would be a good idea to document it here https://docs.btcpayserver.org/FAQ/FAQ-Wallet/#what-is-a-derivation-scheme as well.

Exactly what I was coming back here to say add ! :D
I'll just copy the tip, should do the trick ?

pavlenex:
I hope when we restructure the wallet setup we can change `derivation scheme` in the UI to Extended public key so we can in more detail update that section and replace that term. Just a text would be enough, but if you prefer a tip, why not.

docs/FAQ/FAQ-Wallet.md
```diff
@@ -90,6 +90,10 @@ Using different derivation schemes with your xpub, you can also choose to create
 |Multi-sig P2SH-P2WSH	| 2-of-xpub1...-xpub2...-[p2sh] |
 |Multi-sig P2SH |	2-of-xpub1...-xpub2...-[legacy] |
 
+:::tip
+On top of the Xpub extended public key formats shown above, BTCPay Server supports Ypub and Zpub formats. Please note that these will be converted to Xpub once the wallet setup is completed. This has no effect on how you receive or send funds.
```
pavlenex, line 94: These are usually spelled with lowercase first letter, or all uppercase  so this is for consistency with other sources, haven't seen first uppercase as much compared to the two.

```suggestion
On top of the xPub extended public key formats shown above, BTCPay Server supports yPub and zPub formats. Please note that these will be converted to xPub once the wallet setup is completed. This has no effect on how you receive or send funds.
```

docs/WalletSetup.md
```diff
@@ -11,7 +11,11 @@ There are two ways to set up a wallet in BTCPay Server:
 
 Using an existing wallet assumes that you already have an external wallet created and backed up.
 
-When using an existing wallet with BTCPay Server, you're providing an extended public key (xpub) from an externally generated wallet. Extended public key is a watch-only wallet token. Providing this public key allows BTCPay Server to generate a new address each time a new invoice is created.
+When using an existing wallet with BTCPay Server, you're providing an extended public key from an externally generated wallet. Extended public key is a watch-only wallet token. Providing this public key allows BTCPay Server to generate a new address each time a new invoice is created.
+
+:::tip
+When you input an extended public key starting with Ypub or Zpub as a derivation scheme, BTCPay Server automatically converts it to the Xpub format. This has no incidence on how you receive or send funds.
```
Zaxounette, line 17:
```suggestion
When you input an extended public key starting with yPub or zPub as a derivation scheme, BTCPay Server automatically converts it to the xPub format. This has no incidence on how you receive or send funds.
```

britttttk (approved):

PR #831, opened 2021-01-30
Update Manual Deployment

GraemeConradie:
Line 944 - symbolic link creation fixed - ln -s /etc/bitcoin/bitcoin.conf ~/.bitcoin/bitcoin.conf should be ln -s ~/.bitcoin/bitcoin.conf /etc/bitcoin/bitcoin.conf

### DIFF
--- a/docs/ManualDeploymentExtended.md
+++ b/docs/ManualDeploymentExtended.md
@@ -944,7 +944,7 @@ lnd version 0.10.99-beta commit=clock/v1.0.0-229-ge64e71d86dc1ac716c30a80f85a22e
 lnd looks for bitcoin.conf in a specific location to get necessary RPC and zeromq details.
 
 ```bash
-~$ ln -s /etc/bitcoin/bitcoin.conf ~/.bitcoin/bitcoin.conf
+~$ ln -s ~/.bitcoin/bitcoin.conf /etc/bitcoin/bitcoin.conf
 ```
 
 ##### 4. Create a configuration file.


dennisreimann:
Are you sure about this one? Haven't tested it, but as far as I read it, everything geets installed into `/etc/bitcoin/` and the symlink should be created in `~/.bitcoin/`.

GraemeConradie:
This is a manual deployment and as such you already have a BTC config that sits in ~/bitcoin/bitcoin.conf. This step is where you tell lightning where to find the btc config to get the RPC and zeromq details. The default location for the btc config is ~/.bitcoin/bitcoin.conf. Trying to create a symlink results in an error. lnd looks for bitcoin.conf in a specific location, question is what is that location ?

dennisreimann:
> lnd looks for bitcoin.conf in a specific location, question is what is that location ?

According to the [LND sample config](https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf#L500) which also points to `~/.bitcoin/bitcoin.conf` that seems right.

> Trying to create a symlink results in an error. 

What error are you getting?

GraemeConradie:
You cannot create a symlink to a location that already has a file called bitcoin.conf. My structure, and most people who run local nodes will have the same structure, is that the bitcoin config file lies in ~/.bitcoin/bitcoin.conf.

The manual says : - ln -s /etc/bitcoin/bitcoin.conf ~/.bitcoin/bitcoin.conf // Create a symbolic link from our file in /etc/bitcoin to the location ~/.bitcoin. You cannot. I don't have an issue, but other people will not get past this.

The https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf#L500 you mention is correct, it merely points to ~/.bitcoin. LND will inspect the config file, and use the values it finds there.

britttttk:
@GraemeConradie if you think users may find themselves in both situations you can always add your note wrapped in a [tip text highlight](https://github.com/btcpayserver/btcpayserver-doc#text-highlights). 

Since this is a community doc and you find this to be an improvement then ACK.

britttttk (approved):

pavlenex:
@dennisreimann Can you check the PR but also the failing checks:

```

Run ./setup-deps.sh
Cloning into '/home/runner/work/btcpayserver-doc/btcpayserver-doc/deps/btcpayserver'...
tail: cannot open '66' for reading: No such file or directory
tail: cannot open '72' for reading: No such file or directory
Error: Process completed with exit code 1.
```

dennisreimann:
This is fixed on master, so please rebase (or ignore, because we aren't changing anything link check related here).

Zaxounette:
@GraemeConradie This PR is ready to be merged after a rebase on your part.
Please rebase so that we can get this one done with.

pavlenex:
Should be good enough for merge even without the rebase, as it doesn't touch the files that are out of sync.

PR #832, opened 2021-02-01
Unprivileged user creation

GraemeConradie:
useradd -m (create home), -s (shell) /dev/null is not valid, must be sudo useradd -M (no home) admin && sudo usermod -L (lock account) admin

### DIFF
--- a/docs/ManualDeploymentExtended.md
+++ b/docs/ManualDeploymentExtended.md
@@ -70,7 +70,7 @@ If you subsequently change the iptables rules and want to save them across reboo
 These instructions configure everything to run under an **unprivileged user** called `admin`.  Create this user before proceeding:
 
 ```bash
-~$ sudo useradd -m -s /dev/null admin
+~$ sudo useradd -M admin && sudo usermod -L admin
 ```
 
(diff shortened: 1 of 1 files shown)


Zaxounette:
Pinging the code enabled peeps @dennisreimann and @britttttk .

Is this a correct bash string change ?

dennisreimann:
Why does the account get locked?

GraemeConradie:
> Why does the account get locked?

Service accounts don't have a "proper" login shell, i.e. they have /usr/sbin/nologin as login shell (or, back in the old days, /bin/false). Service accounts should be locked, i.e. it is not possible to login (for traditional /etc/passwd and /etc/shadow).

This can be achieved in different ways, MatterMost suggests : sudo useradd --system --user-group mattermost, RocketChat on the otherhand uses : sudo useradd -M rocketchat && sudo usermod -L rocketchat.

I prefer -M -L, and assigning ownership to the required directories and files.

dennisreimann:
Ok, thanks for elaborating, I learned something today! 👍

dennisreimann (approved):
ACK.

PR #843, opened 2021-02-16
update manual deployment with testnet commands

rukykf:
I recently had to do a minimal setup of BTCPayServer in testnet mode and it took me a while to find the testnet commands to do it so I thought these additions to the docs might be helpful.

### DIFF
--- a/docs/ManualDeployment.md
+++ b/docs/ManualDeployment.md
@@ -30,6 +30,7 @@ The docker deployment will provide you easy update system and make sure that all
 
 This steps have been done on Ubuntu 18.04, adapt for your own install.
 
+For Testnet specific deployment, after installing Bitcoin, .NET Core, NBXplorer and BTCPayServer, see [Commands for Running in Testnet Mode](#testnet-specific-deployments)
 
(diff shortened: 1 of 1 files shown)


Zaxounette (changes requested):
Hi @rukykf ! 

Thanks for this addition.

On the wording and formatting side, to separate explicitly mainnet and testnet deployments, it would be preferable to
* Add these commands as a seperate H2 title, right before "Additionnal links".
* Include a simple line at the beginning of the file with a wording in the lines of `For Testnet specific deployment, see [LINK TO BOTTOM OF PAGE WHERE YOU COMMANDS ARE].` 

As for the actual code, i'll leave that to @britttttk .

rukykf:
Thanks for your feedback @Zaxounette. I just made the changes you requested.

Zaxounette (approved):
ACK for wording and formatting, thanks for the rapid changes =)

FYI, we have a ContributeDevTest.md file that relates to testing BTCPay, but in regtest and locally only.
Britt might want to add these commands to that particular file instead.
Will wait on her input.

britttttk (approved):

PR #856, opened 2021-02-26
Hardware As A Service: Add Umbrel, remove LuxNode

dennisreimann:
LuxNode doesn't seem to be available (anymore?), the URL does not respond and I couldn't find any information on that service online.


Zaxounette (approved):
Went and found their twitter account.
It's long dead and populated by a follow bot.

ACK.

PR #860, opened 2021-03-01
HardWareAsAService.md support disclaimer

### LINKED ISSUE #853: Add info to docs about using BTCPay with other services
A lot of comments in the chat come from users with BTCPay on things like MyNode, Umbrel, Raspiblitz, etc. We should mention in docs that all of these are different ways of implementing BTCPay and we are not able to answer questions about those setups. 

Discussed in [this docs meeting](https://github.com/btcpayserver/organization/issues/23). Talked about adding this disclaimer to Hardware As a Service document.

Zaxounette:
closes #853

* Adds a disclaimer in HardwareAsAService.md stating that questions, bugs and issues should be directed to providers.

### DIFF
--- a/docs/HardwareAsAService.md
+++ b/docs/HardwareAsAService.md
@@ -16,4 +16,8 @@ Here you can find a non-exhaustive list of companies that provide such a service
 - [RaspiBlitz](https://shop.fulmo.org/raspiblitz/)
 - [Umbrel](https://getumbrel.com/)
 
+:: Warning
+BTCPay Server cannot answer questions relating to integrations made by third-party entities. Please report bugs and issues directly to the entity that provided you with BTCPay Server inside their product.
+::
+
 Do you provide Hardware As A Service and are not listed here ? Open an issue to [get added to this list](https://github.com/btcpayserver/btcpayserver-doc/issues)
--- a/docs/HardwareAsAService.md
+++ b/docs/HardwareAsAService.md
@@ -16,7 +16,7 @@ Here you can find a non-exhaustive list of companies that provide such a service
 - [RaspiBlitz](https://shop.fulmo.org/raspiblitz/)
 - [Umbrel](https://getumbrel.com/)
 
-:::danger
+:::warning
 Hardware-as-a-Service products consist of multiple softwares tied together.
 The BTCPay Server community has no knowledge of how BTCPay Server integrates into these services, and cannot provide any support for these types of [Manual Deployments](ManualDeployment.md). Please report bugs and issues directly to the entity that provided you with BTCPay Server inside their product.
 :::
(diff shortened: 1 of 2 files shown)


britttttk (commented):
Maybe we can link to this https://docs.btcpayserver.org/Troubleshooting/#4-asking-for-help document somehow or point them to the troubleshooting doc... 

 It would be great if we could connect the idea that these are essentially "custom deployments" of btcpay and the community does not provide extensive support for these types of deployments (like the warning in the link).

Not sure how to word that... what do you think @Zaxounette

Zaxounette:
I feel like pointing them to `asking for help` is giving them the stick to beat us, instead of the HAAS providers.

As for the warning, I might just as well copy it entirely and add it on top of the other.

dennisreimann:
I agree with @Zaxounette. We should be consistent when we say we don't offer support for external services – pointing to the Troubleshooting page kind of contradicts that. 

Also think that adding such a disclaimer to the Troubleshooting page makes sense. 👍

Zaxounette:
Okay so:

* 0171f06 did a second pass on the HAAS.md disclaimer to refine it a bit.
* 0669502 Added a sentence in troubleshooting.md at the end of the already existing disclaimer indicating explicitly that it includes HAAS products.

docs/HardwareAsAService.md
```diff
@@ -16,8 +16,9 @@ Here you can find a non-exhaustive list of companies that provide such a service
 - [RaspiBlitz](https://shop.fulmo.org/raspiblitz/)
 - [Umbrel](https://getumbrel.com/)
 
-:: Warning
-BTCPay Server cannot answer questions relating to integrations made by third-party entities. Please report bugs and issues directly to the entity that provided you with BTCPay Server inside their product.
-::
+:::danger
```
pavlenex, line 19: Please do not use `dangerous` unless something is really dangerous. In this case a blue mark is absolutely fine. If we keep using dangerous often, people may ignore it.

Zaxounette: Good point.

By blue mark, you mean :::tip ?

pavlenex: Personally, I don't think it's a warning either. It's simply a prominent indicator that officially something isn't supported.

dennisreimann (commented):
Pav's note about the type is good, other than that it LGTM.

britttttk (approved):

britttttk:
Nice one @Zaxounette much better explanation

Zaxounette:
After discussion with Dennis, 6127a81 pushes the disclaimer up above fold for early view by the user.

Also adds a title "please note" to the container.

ACK.

PR #862, opened 2021-03-04
Improve internal link rewrite

dennisreimann:


Zaxounette (approved):
ACK on the link change in FAQ-general.

As for the actual code part, i'll leave that to the code junkies out there :)

dennisreimann:
For the code changes: As long as the link check passes and the links that were opening new tabs are now working internally, everything should be good. :)

PR #865, opened 2021-03-08
SEO: Improve custom properties

dennisreimann:
This adds some good SEO defaults to our docs and allows to customize the description and keywords/tags for each page individually. I've added some properties for the Apps and WooCommerce pages as an example.

In case a page does not have those properties, the keyword "BTCPay Server" is used and the first paragraph becomees the description.

Adds general SEO properties to the site, so that we now also have a good looking card view for doc URLs that get shared.


pavlenex (approved):
Good improvement.

Should be added to other pages as well in the future, we should open an issue.
Could be something @RiccardoMasutti may be interested in.

dennisreimann:
Yes, now that this is merged we can go through the pages and update them where we see fit.

Issue #855, opened 2021-02-26
Fix 'what if I have an issue paying an invoice' FAQ


britttttk:
[FAQ Question](https://docs.btcpayserver.org/FAQ/FAQ-General/#what-if-i-have-a-problem-paying-an-invoice)

This page tells them to edit the wallet and add a -[p2sh] to the end, but we need to explain that this means it will become an entirely new wallet. 

- If they are using it as a hot wallet, their seed won't be stored with the change
- If they imported it from another wallet, their wallet wont track these payments and the addresses generated
- Old funds wont be available in the wallet feature

[Comment from chat](https://chat.btcpayserver.org/btcpayserver/pl/37s8k8k65jyj8fbc6jfgwgkqgw)

Zaxounette:
In addition to this info, we would need more:

The seed isn't stored with the change: What does that mean ? 
It becomes a hot wallet ? It becomes a "seedless" address ? Else ?

Kukks:
> In addition to this info, we would need more:
> 
> The seed isn't stored with the change: What does that mean ?
If you generated the wallet using the generate wallet option and ticked the "Is Hot wallet" or the "import to rpc" feature, they will now no longer be enabled.

PR #869, opened 2021-03-11
Updates to wallets FAQ

### LINKED ISSUE #838: Wallet Redesign FAQ Updates
Review the [Wallet FAQ questions](https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/FAQ/FAQ-Wallet.md) to make sure they are still relevant or need to be improved with respect to the wallet redesign. I think the first few questions could use an update.

Relates to #837

britttttk:
Close #838

### DIFF
--- a/docs/FAQ/FAQ-Wallet.md
+++ b/docs/FAQ/FAQ-Wallet.md
@@ -18,23 +18,23 @@ This document contains frequently asked questions related to BTCPay Server's [in
 
 BTCPay Server has an internal wallet which you can use to view incoming and outgoing associated Bitcoin transactions and spend your funds.
 
-It works like any other wallet, but has enhanced privacy features (non-custodial, no third-parties, verified with a dedicated full node, etc.) by default and also solves certain UX problems you may encounter when [using an existing wallet] with BTCPay Server.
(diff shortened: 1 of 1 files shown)


Zaxounette (approved):

docs/FAQ/FAQ-Wallet.md
```diff
@@ -96,7 +96,7 @@ Using different derivation schemes with your extended public key, you can also c
 |Multi-sig P2SH |	2-of-xpub1...-xpub2...-[legacy] |
 
 :::tip
-On top of the xPub extended public key formats shown above, BTCPay Server supports yPub and zPub formats. Please note that these will be converted to xPub once the wallet setup is completed. This has no effect on how you receive or send funds.
```
britttttk, line 99: Removed this because it is too likely to confuse users that the xpub doesn't really matter. Although the z/x/yPub doesn't matter, it is quite common for users to just arbitrarily modify their xpub thinking it doesn't really matter. (Examples like issue [#855])

Zaxounette: This tip was initially put because of the recurring questions about "I put my public key, but it got changed, have I been hacked ?" where a zpub or y pub was added and user didn't know it was converted to xpub.

britttttk: @Zaxounette ok I have added it back that is also a fair point and user question. Thanks!

dennisreimann (approved):

PR #882, opened 2021-04-01
Change HTTP to HTTPS links

l33d4n:
I changed some of the HTTP links to HTTPS to avoid redirects and improve their loading time.


dennisreimann (approved):

Kukks (approved):

dennisreimann:
The links that fail the linkcheck aren't exactly broken, but maybe those sites are blocking requests coming from our pipeline. We'll need to keep an eye on that and potentially need to add these hosts to the exclude list if it persists.

PR #895, opened 2021-04-11
Docs/apps.md clarification

### LINKED ISSUE #894: Improve BTCPay Server Apps
I could not find the information I was looking for on the "BTCPay Server Apps" page (`/Apps/`).

[PLEASE DESCRIBE HOW THE PAGE CAN BE IMPROVED]

I would like to use BTCPay for tips. So the person leaving the tip would decide how much to tip me.

Is this possible?

Zaxounette:
Closes #894 

In a question brought up in #894 it appeared that we didn't mention explicitly `tips` as being a possible use-case of the `POS app`.
Since the `standalone donation button` is probably going to be a gonner in the future, we might as well mention it for the `POS app`.

### DIFF
--- a/docs/Apps.md
+++ b/docs/Apps.md
@@ -21,7 +21,7 @@ The **web-based PoS app** allows users with brick and mortar stores to readily *
 
 Adding new products is easy. The app has a **shopping cart feature**, **tips**, **product inventory**, **custom payment options** and more.
 
-**Point of sale app** can also be used to receive donations, tips or even be used as a small e-shop, depending on the options or customizations applied.
+The **Point of sale app** can also be used to receive donations, tips or even be used as a small e-shop, depending on the options or customizations applied.
 
 Curently, the **Point of Sale app** supports three different views:
 * A `Static` view representing only the items for sale.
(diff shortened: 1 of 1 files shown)


docs/Apps.md
```diff
@@ -21,7 +21,7 @@ The **web-based PoS app** allows users with brick and mortar stores to readily *
 
 Adding new products is easy. The app has a **shopping cart feature**, **tips**, **product inventory**, **custom payment options** and more.
 
-**Point of sale app** can also be used for receiving donations or even as a small e-shop, depending on the customizations applied.
+The **Point of sale app** can also be used to receive donations, tips or even be used as a small e-shop, depending on the options or customizations applied.
```
Kukks, line 24: hmm I wonder if you would say e-shop or just "e-commerce usage"

Kukks (changes requested):

docs/Apps.md
Zaxounette:
```suggestion
The **Point of sale app** can also be used to receive donations, tips or even be used as a small e-commerce shop, depending on the options or customizations applied.
```

You're right, e-shop doesn't sound right.
How's the above ?

PR #898, opened 2021-04-24
Wallet provider docs updates

### LINKED ISSUE #881: Updates to wallet provider documentation
Relates to #837 

Need to make sure all the wallet provider docs we have are also updated with the new wallet ui redesigns. 
Examples: Cold Card, Electrum, Wasabi, etc.

https://docs.btcpayserver.org/ColdCardWallet/#connecting-coldcard-wallet-to-btcpay-server
Maybe could use new videos if someone is feeling up to it :)

britttttk:
close #881 

- electrum
- wasabi
- cold card


britttttk:
@Zaxounette what do you think about the gifs? I just wanted to give it a try but it's a little confusing because they loop... I was thinking about removing them.

Do we still want to keep the xpub method in the doc? Or is it too confusing calling it "alternative method".

They can disable encryption and re-enable it but do we want to show that? Since we are doing new wallet... it could be confusing.

Zaxounette (approved):
* I'm all for gifs, less images and a less lengthy page.

* xpub is an alternative, so wording it like that is fine IMO.

* People that use electrum with encryption should already know that an encrypted file cannot be read by anyone that doesn't have the keys. Otherwise they should go back to basics. Keeping it like this is fine as well.

ACK the gifs,
ACK the text changes ! =)

docs/WasabiWallet.md
```diff
@@ -50,6 +50,9 @@ Select and **copy** the `Extended Account Public Key`. This is the **public** ke
 
 ![Connect Wasabi Wallet to BTCPay Server](./img/Wasabi/WasabiWalletSetupBTCPay8.png "Connect Wasabi Wallet to BTCPay Server")
 
+1. Return to your BTCPay Server. Click on the `Stores` in the header menu
+2. Click Settings > Setup > Connect an existing wallet > `Enter extended public key`
+
```
kevinrichard669, line 55: BestDnT5150@gmail.com

kevinrichard669 (commented):
[BestDnT5150@gmail.com ](http//:BestDnT)

docs/WasabiWallet.md
kevinrichard669: Definitely 👍  worth checking out!!

Kevinho5150:
> @Zaxounette what do you think about the gifs? I just wanted to give it a try but it's a little confusing because they loop... I was thinking removing them for photos.
> 
> Do we still want to keep the xpub method in the doc? Or is it too confusing calling it "alternative method".
> 
> They can disable encryption and re-enable it but do we want to show that? Since we are doing new wallet... it could be confusing.

PR #902, opened 2021-05-02
Add resources for getting help with c-lightning

niftynei:


britttttk (approved):

pavlenex:
Why did you remove the IRC? Is it no longer active? Do we have someone working on c-lightning to confirm this is a legit group? Want to make sure we're not just being used for promoting an unofficial telegarm group.

pavlenex (approved):

PR #909, opened 2021-05-26
Update FAQ-Altcoin.md

FeliciaAKelley:


Zaxounette (changes requested):
I'm pretty sure this is an error. 
Please confirm, or explain what you're looking to do with these changes.

pavlenex:
Spam.

FeliciaAKelley:
> I'm pretty sure this is an error.
> Please confirm, or explain what you're looking to do with these changes.





There is one I do have that access to the btcpayserver from my account in Google however I am unable to verify it because I'm unable to verify my domain which makes no sense because I did have a g Suite account so something in the data somewhere Iwas tampered with or expense account takeover I switched it out whatever you can think of I don't know or I'll just be the work of jealous cuz I don't know one of the two that's all I see when I look at it

PR #920, opened 2021-06-14
Add new supporter pnxbet to readme

pavlenex:


Zaxounette (approved):

pavlenex:
Hm test is failing on pnxbet probably because of their firewall giving 403

Zaxounette:
> Hm test is failing on pnxbet probably because of their firewall giving 403

I can confirm that as well when looking into the linkcheck.
@dennisreimann This might be a good candidate to exclude from the linkcheck since they have an IP filter for legal reasons.

dennisreimann:
Opened #922 to fix that. Thanks for the pointer.

Issue #917, opened 2021-06-14
Custom reverse proxy to Tor not working


danielcharrua:
Following the docs guide on creating a [reverse proxy to Tor](https://docs.btcpayserver.org/ReverseProxyToTor/) I get a blank page and error 400 after sending the login information. I hace strictly followed all steps mounting this on a DigitalOcean fresh droplet.

I've found that commenting out `proxy_set_header Connection "upgrade";` I can login to BTCPay Server but with the following errors on the JS console
```
WebSocket connection to 'wss://pay.xxx.es/Notifications/SubscribeUpdates' failed: Unexpected response code: 400
Error while connecting to websocket for notifications (callback)
```

dennisreimann:
Does it work fine in Tor browser using the address provided as `REMOTE_HOST`?

danielcharrua:
> Does it work fine in Tor browser using the address provided as `REMOTE_HOST`?

Yes, just tried it on a Tor browser and I can login without any issues and any console warnings/errors

dennisreimann:
Do the nginx logs on both the remote host and the proxy provide any more information?

dennisreimann:
> I can't see anything else. Has anyone have tried this setup?

I have it running in production and wrote the doc based on my setup. The connection upgrade is for the web sockets, but shouldn't lead to a general error 400.

danielcharrua:
I see lots of people struggling with this in particular but any of them have a solution or workaround. I understand that the connection upgrade is for web sockets but in my case is leading to a 400 error and a blank page is shown. I have followed your steps and started again 3 times 😬

I have the BTCPay Server installed in a Raspi, under Umbrel. Maybe my router is in the way? This is possible?
The weird thing is that if I use a Tor browser all goes smooth... Is there anything else I can test to help debugging this?

dennisreimann:
The router shouldn't be involved here, this is one of the main points of this setup. 
Did you try the reverse SSH as an alternative?

danielcharrua:
Reverse SSH is not an alternative at this moment as I don't want to install any extra config on the Umbrel node. But I think I've found a way, just checking I've found this [http://nginx.org/en/docs/http/websocket.html](http://nginx.org/en/docs/http/websocket.html) and added to `/etc/nginx/conf.d/map.conf`

```
map $http_upgrade $connection_upgrade {
  default upgrade;
  ''      close;
}
```

[...]

PR #923, opened 2021-06-15
Updated mapping and proxy requests

danielcharrua:
Solved WebSockets proxy issues.


dennisreimann (approved):
ACK, as discussed in #917.

PR #930, opened 2021-06-18
Basic info page about API keys

woutersamaey:
Comes in handy when we need to explain API keys and where to find them.
Some apps that need an API key can link to this page.


Zaxounette:
I feel like expanding on an already existing doc would be best for this paragraphe.

[This doc](https://docs.btcpayserver.org/GreenFieldExample/#create-a-new-api-key) covers the basics (and an example) of the Greenfield API.
Your paragraphe could be integrated in it.

Changing the actual title of that page could make sense as well, to reflects the added information.

woutersamaey:
@Zaxounette I don't mind moving the content, only the page you're mentioning is a technical page (it talks about cURL), whereas the page I'm thinking of would be for users who don't know what APIs are. They'd only know they need to go somewhere, click a few buttons and permissions and copy the key.

Zapier wants a page like that in your docs so it can be referenced to users as they're trying to figure out what an "API Key" even means.

Zaxounette:
Your comment makes sense.

That being said, API key settings are nested inside the user settings.

We actually have no page relating to those options (password, FIDO/2FA, API keys, etc...) under user settings.
I propose that you nest your document under `FAQ and common issues` in the config.js file, and inside the FAQ folder. `FAQ-UserSettings.md` could be an appropriate name.

That way you link to your section on that page in Zapier.

If you want to add the basic info for the rest of the user settings, go ahead, otherwise i'll do it once your PR is merged.

britttttk:
> expanding on an already existing doc would be best for this paragraphe

agree, this does not seem to need a new doc, adding to FAQ seems better

dennisreimann:
@woutersamaey Can you move it to the FAQ as suggested?

pavlenex:
ping @woutersamaey what do we want to do with this one? Is it still relevant?

woutersamaey:
@pavlenex yeah we're still going to need it for Zapier's docs. They desire a doc page that explains API keys... I'll rededicate myself to this task.

woutersamaey:
Not needed anymore.

PR #956, opened 2021-08-19
Fix for mobile layout on firefox

kngako:
related to the vuepress issue with the mobile layout removing viewport meta tags on firefox for Android.

Fix is to explicity define it.

https://github.com/vuejs/vuepress/issues/2369


dennisreimann:
Thanks for bringing this up and fixing it!

PR #957, opened 2021-08-19
try to fix link test (test)

pavlenex:


pavlenex:
ok so blockstream blocks our test for sure

PR #984, opened 2021-11-07
Rpi4 install

lightninginabox:


pavlenex:
@lightninginabox I assume this is just an update, long-term we should delete RPI3 page and use instructions for the image you've made. Should I merge this one meanwhile?

PR #986, opened 2021-11-16
include Althash in list of available altcoins

denuoweb:


pavlenex:
I don't think this coin is integrated?

pavlenex (approved):

PR #991, opened 2021-12-16
Zapier integration docs

woutersamaey:


pavlenex:
I think we can just pull readme from Zapier repo, just like we do for docker, transmuter, etc, no need to add an .md?

pavlenex:
This should be moved to readme and we should then import it in the docs as we do with all other plugins. Maybe @dennisreimann can give you direction on how to do it, as it's less of a maintenance burden.

The docs will still be in btcpay docs, they'd just be pulled from the readme of your Zapier repo and maintained there.

woutersamaey:
Oh ok, that works too...

pavlenex:
If you prefer to keep them separate, and completely different, then that's fine as well but in case of an overlap we can host them in a readme. Whatever is a consensus is fine by me. I believe @britttttk and @Zaxounette will review it.

woutersamaey:
I found the script that merges the docs, so no need to bother @dennisreimann with this.

I am warming up to your idea and doing the README in the Zapier repo + auto-merging that. I can propose the changes to the script as well.

Should we close this PR then?

Zaxounette:
You can close this PR if you're using the zapier repo readme.

PR #993, opened 2021-12-20
Beautify docs/img/Architecture.png

rene78:
Just some minor visual changes to diagram.


dennisreimann:
Thanks, looks better now! 

As GitHub does not show the diff, here's the before and after version:

PR #1005, opened 2022-01-19
Fix “ssl” Directive Is Deprecated

SomeBTChomer:
Fix “ssl” Directive Is Deprecated, Use “listen … ssl”

The deprecation warning tells you to reconfigure your SSL settings. In nginx 1.10 (and below) you configured SSL using the ssl on;
This setting changed in nginx 1.12 (and above). You now need to configure SSL in the same line as the listen statement. 

Simple fix. Remove the "ssl on" completely

Thanks for the otherwise great tutorial, everything worked out fine after this simple fix!

### DIFF
--- a/docs/Deployment/ReverseProxyToTor.md
+++ b/docs/Deployment/ReverseProxyToTor.md
@@ -171,7 +171,6 @@ server {
   server_name mydomain.com;
 
   # SSL settings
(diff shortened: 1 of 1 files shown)


dennisreimann (approved):
Thanks for the contribution!

PR #1024, opened 2022-02-17
Adding troubleshooting section and updating WooCommerce FAQ.

ndeet:


pavlenex (approved):
There are probably a few things to iron out, but let's see if we can find contributors that can help us improve the copy here and on the wordpress.org for maximum SEO impact and increased reach.

PR #1048, opened 2022-03-23
Update UseCase.md

### LINKED ISSUE #996: Revise the need and positioning of the `Use-Case` page.
We should discuss and determine if `Use case` page is relevant to the users. If so, we also should determine If we want it to be second position page in the sidebar navigation. Perhaps pages like `Try it out` could be more prominently displayed.

- [x] Determine if we need the `Use case` page
- [x] Revise the positioning of the `Use case` page in the sidebar
- [x] Revise the content make sure it's clear and concise

pavlenex:
Closes #996

Note: squash before merging

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -19,13 +19,13 @@ const extractDescription = text => {
 
 const sidebarUserGuide = [
   {
-    title: 'Basics',
+    title: 'Learn',
     collapsable: false,
     children: [
       ['/Guide', 'Introduction'],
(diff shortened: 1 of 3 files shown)


docs/.vuepress/config.js
```diff
@@ -19,13 +19,13 @@ const extractDescription = text => {
 
 const sidebarUserGuide = [
   {
-    title: 'Basics',
```
pavlenex, line 22: @dennisreimann I hope this won't break anything 😄

PR #1060, opened 2022-03-27
Minor Update DevCode.md

Bangalisch:
Hey, 

Found a little textual thingy in the Guide. 
A double Now :) 

Also added another ttxt change to it, as the views have changed on latest update. 
The old Welcome view is no longer, it shows the dashboard as entry. 
Changed the path for this first visual example to change and added the img at the right location. 

Edit on the GreenFieldExample.md minor textual fixes, reads a little easier. 

Sorry if there's other guidelines to this. 
I'll do better next time. 

Cheers.


dennisreimann (approved):
Thanks!

PR #1064, opened 2022-03-29
Add border around content images

### LINKED ISSUE #1057: Add minimal boarder around images
Our documentation mostly contains screenshots of the software. And since we're such an awesome group that respects design guidelines we've made, we use same background colors. For certain images, that means it's very hard to distinqush them.

An example:
Screenshot 2022-03-24 at 21 48 57

I am not sure of what would work best, but I propose we add either a boarder or drop a shadow, @dennisreimann any ideas on this one?

dennisreimann:
Closes #1057.


pavlenex (approved):
tACK, thanks Dennis!

Issue #998, opened 2021-12-22
Improve `Try it Out` page


pavlenex:
As an important page where user should understand how quickly to give BTCPay a try, it's critical that we make sure the page provides such information.

At the moment, it's very confusing and structured poorly.

- [ ] Revise the content
- [ ] Remove introduction to btcpay video as it's not really relevant to the page
- [ ] Consider a new video that may showcase people how to try btcpay in a few ways (mainnet demo, third-party host, local dev, etc)

pavlenex:
#1066

PR #1066, opened 2022-03-29
PR for issue 998 - only missing new video content

Bangalisch:
Just needs new video content.
Moved Testnet away from this page as a avg user will not use it. 
It was moved to Deployment, it could use a little touching up with Screenshots.


Bangalisch:
#998

pavlenex (changes requested):
Nice idea, agreed with it, but I think `Developers` is better suitable place for this then deployment. Dev are more likely to use testnet to test things out then people trying to find docs on how to actually deploy btcpay, though the line is thin there, I suggest we add it in `Developers` docs, specifically here:

Screenshot 2022-03-30 at 11 06 11

What do you say?

Bangalisch:
Agreed! 
Wasnt to sure about the Testnet stuff where to put it, but ur right! 
Lets move it there

docs/.vuepress/config.js
```diff
@@ -23,10 +23,10 @@ const sidebarUserGuide = [
     collapsable: false,
     children: [
       ['/Guide', 'Introduction'],
+      ['/TryItOut', 'Try it out']
```
dennisreimann, line 26:
```suggestion
      ['/TryItOut', 'Try it out'],
```

PR #1103, opened 2022-05-03
Consolidate Raspberry Pi Deployment docs

### LINKED ISSUE #467: Consolidate RPI documentation
- [ ] Add disclaimer in RPI3.md that it's not mean for production due to stability issues and lack of RAM.
- [ ] Improve RPI4 docs and standardize them, as it's hard to follow up and one as to look at both RPI3 and RPI4 docs to deploy things on RPI.
- [ ] Suggest RPI Imager and Balena etcher instead of CLI method

dennisreimann:
This consolidates the existing Raspberri Pi deployment docs and updates the instructions for the RPI 4 setup:

- Use official Imager app to set up the SD Card, because it allows to configure the necessary options via UI instead of the CLI.
- Take the FastSync part from RPI 3 docs, otherwise discard them as the RPI 3 should not be used anymore. 
- Update and generalize the install script to allow for configuring the Lightning node and additional options.

Closes #467.

Went through the steps and set up a new testing instance for myself. Works like a charm - next potential optimization would be to leverage the [BTCPi](https://github.com/lightninginabox/btcpi) setup once it is done.


docs/Deployment/RaspberryPi4.md
```diff
@@ -0,0 +1,328 @@
...
+```bash
+apt install -y ufw
+ufw default deny incoming
+ufw default allow outgoing
+```
+
+This command allows SSH connections from internal networks only:
+
+```bash
+ufw allow from 10.0.0.0/8 to any port 22 proto tcp
```
gradientskier, line 238: You may consider to open port 53 too, for those who will enable pihole. 
I use eth0 and wlan0 to filter on the local network, but I had to rename the docker bridged interface as br-btcpayserver otherwise the ssh connections from btcpayserver container wouldn't work.
See my ufw configuration at https://github.com/gradientskier/btcpayserver-docker/blob/arm64-bullseye-prod/Ansible/roles/ufw/templates/expected_ufw_output.txt.tmpl

docs/Deployment/RaspberryPi4.md
```diff
@@ -0,0 +1,328 @@
...
+A 1TB SSD allows you to keep a full copy of the Bitcoin blockchain.
+You can also use BTCPay Server without a full copy of the Bitcoin blockchain by using the [pruning option](/Docker/#how-i-can-prune-my-nodes).
+
+### Power Adapter options
+
+- [Official Raspberry Pi 4 USB-C Power Adapter 5.1V/3A for US](https://shop.pimoroni.com/products/raspberry-pi-official-usb-c-power-supply-us?variant=29391144648787) ($10)
+- [Official Raspberry Pi 4 USB-C Power Adapter 5.1V/3A for EU](https://shop.pimoroni.com/products/raspberry-pi-official-usb-c-power-supply-eu?variant=29391130624083) ($10)
+- [Official Raspberry Pi 4 USB-C Power Adapter 5.1V/3A for AU](https://shop.pimoroni.com/products/raspberry-pi-official-usb-c-power-supply-au?variant=29391160737875) ($10)
+
+Don’t waste your time with random Chinese power adapters from Amazon, or expect that the existing ones you have at home are going to work fine. The Raspberry Pi 4 has issues with unofficial adapters, and for only $10 it’s better to just **get an official adapter** instead of learning this the hard way.
```
Bangalisch, line 40:
```suggestion
Don’t waste your time with random cheap power adapters from Amazon, or expect that the existing ones you have at home are going to work fine. The Raspberry Pi 4 has issues with unofficial adapters, and for only $10 it’s better to just **get an official adapter** instead of learning this the hard way.
```

dennisreimann: Thanks!

docs/Deployment/RaspberryPi4.md
dennisreimann: As the Docker fragments are optional and additional, I would say we should focus on the minimum here and rather document this as an additional step in the [Pi-Hole docs](https://github.com/btcpayserver/btcpayserver-docker/edit/master/docs/pihole.md). Can you add it there?

pavlenex (approved):
LGTM

PR #1109, opened 2022-05-11
Update to the Dreamlab web deployment. 

Bangalisch:
i missed a small part in the guide. 
I've added the VPS management page and a small description of where to find it.


pavlenex:
@Bangalisch this seems like a conflict, it seems you did a change on a branch that's out of sync and now it added all the changes all over, I think the easiest approach here is to close a PR, make sure your master is in sync iwth the the upstream master and create a new branch with the changes.

PR #1132, opened 2022-06-05
Update WooCommerce Troubleshooting

danielcharrua:
This adds a "success" response to docs for others to see that a 500 response is OK and the request is not being blocked with a 403.


ndeet (approved):
Makes sense, thank you!

PR #1136, opened 2022-06-12
Update Hardware as a Service page

AaronDewes:
This updates the Umbrel link to their latest website.

Also, a warning should be added for Umbrel. A lot of people are running a shop on Umbrel without realizing Umbrel prohibts commercial use of their software. Using an Umbrel node to sell anything is commercial use.

### DIFF
--- a/docs/Deployment/HardwareAsAService.md
+++ b/docs/Deployment/HardwareAsAService.md
@@ -19,7 +19,7 @@ Here you can find a non-exhaustive list of companies that provide such a service
 - [Hack0](https://www.dglab.com/en/works/hack0/)
 - [LightningInABox](https://lightninginabox.co/)
 - [MyNode](https://mynodebtc.com/)
-- [RaspiBlitz](https://shop.fulmo.org/raspiblitz/)
+- [RaspiBlitz](https://raspiblitz.org/)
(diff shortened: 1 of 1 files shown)


pavlenex:
@lukechilds is this true? I just don't see how selling products via btcpay violates Umbrel's license?

AaronDewes:
According to their license, Umbrel can only be used for noncommercial purposes, selling something clearly is commercial. Maybe I'm misinterpreting their license though, I'm not a lawyer.

lukechilds:
Nah, this isn’t something that’s prohibited. So all good. Here’s the license [FAQ](https://github.com/getumbrel/umbrel/wiki/License-FAQ).

pavlenex:
Thanks for clarifying @lukechilds.
@AaronDewes how about we remove the warning and add a new link then? Also we can add runcitadel there right? I don't think we reference to it but you can run btcpay on it, correct?

AaronDewes:
I also noticed the RaspiBlitz link leads to the Fulmo shop, which sells prebuilt RaspiBlitz nodes. Maybe https://raspiblitz.org/ would be a better link?

pavlenex (approved):

Issue #774, opened 2020-11-22
Document adding Webhooks


britttttk:
Document the new feature: create webhooks with BTCPay which is being added to the next major release.

![image](https://user-images.githubusercontent.com/39231115/99898786-64d40b00-2c61-11eb-9d2a-70d1bd9e71b1.png)

Stores > Settings > Webhooks > Create a new webhook

![image](https://user-images.githubusercontent.com/39231115/99898809-9d73e480-2c61-11eb-8e50-49860d76604e.png)

pavlenex:
#1140

PR #1140, opened 2022-06-14
Webhooks update to FAQ

Bangalisch:
Yo,

Saw @britttttk her post #774 :) 
Searched a little, and mostly we tell at specific implementations to do it (see VirtueMart). 
Added a small description to the FAQ about webhooks


petzsch:
Perhaps we can add some example code like the one from the php greenfield client as well? It always helps me with custom integrations to get this working: https://github.com/btcpayserver/btcpayserver-greenfield-php/blob/master/examples/webhook.php

Bangalisch:
I messed up @pavlenex shud be fixed :) 


> Perhaps we can add some example code like the one from the php greenfield client as well? It always helps me with custom integrations to get this working: https://github.com/btcpayserver/btcpayserver-greenfield-php/blob/master/examples/webhook.php
> 
> **EDIT:** Might also wanna explain the different types of webhooks, show an example payload(might be to much for a FAQ entry) and explain the difference over the old NotificationURL from the Bitpay compatible API.

So in that case we'd better pour that out into a specific page dedicated to Webhooks. 
Dont close the britt issue yet, if we want a dedicated PR for a full page on Webhooks, we'll see this in FAQ as a nice to have extra then :)

pavlenex (approved):

PR #1141, opened 2022-06-21
Update to introduction img

Bangalisch:
Update to introduction img.
Looks much better then the old i did. Clean up aswell in this PR if its better.


Bangalisch:
@dstrukt Could you maybe review the img ? I feel like this is much better then the original.
Same assets but swapped to tea shop instead of wallet cuz more pleasing to the eye.

PR #1148, opened 2022-07-01
Expose to clearnet through Cloudflare argo tunnel

Bangalisch:
Made a first write-up of the expose to clearnet via Cloudflare PR https://github.com/btcpayserver/btcpayserver-docker/pull/667

cc @NicolasDorier @danielcharrua for review and additions and maybe relocate it? 

Closes #1147

PR #1149, opened 2022-07-03
Re-arrange the hardware documentation

NicolasDorier:
Ping @Bangalisch, @danielcharrua, @dennisreimann 

This is a follow up of https://github.com/btcpayserver/btcpayserver-doc/pull/1148 .

1. I have moved the documentation of cloudflare to the hardware category, as there is no real advantage to use it elsewhere.
2. I renamed the sections concerning solutions to expose your server on internet
3. Changed `Hardware.md` so it isn't an introduction of a specific hardware deployment, and instead an introduction to the three solutions we are documenting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.vscode
 /deps
 /docs/.vuepress/dist
 /node_modules
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -158,6 +158,7 @@ const sidebarDeployment = [
       {
         title: 'Hardware Deployment',
         path: '/Deployment/Hardware',
+        collapsable: false,
         children: [
           {
             title: 'Raspberry Pi Deployment',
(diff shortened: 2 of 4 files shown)


dennisreimann:
I think it's a bit unfortunate that the categorization now initially hides e.g. the Raspberry Pi deployment, which people might be looking for. We could set [collapsible](https://vuepress.vuejs.org/theme/default-theme-config.html#sidebar-groups) to false to circumvent that issue.

Also: Shouldn't we group the web/cloud deployments like this as well?

NicolasDorier:
Yes I think we should, but in another PR.

Myself I only know Lunanode, whereas there are quite a lot of options nowadays I'd like we have at a single place documented.

dennisreimann (approved):

dennisreimann:
Yup, works for me!

I'll move the Azure and Google Cloud docs into Web Deployment in a separate PR tomorrow.

danielcharrua:
great job guys! looks much better to me now

PR #1159, opened 2022-08-28
Update Wallet.md

PsklQc:
Proposing to add Sparrow has a wallet that allows gap-limit configuration. Although then you might have to add "Connecting Sparrow wallet to BTCPay Server...


pavlenex (changes requested):
The links are pointing to a non-existing pages. If you'd want this added in this way, you'd have to create SparrowWallet.md add content to it, and then do the same for FAQ.

Issue #950, opened 2021-08-12
Update hot wallet signing considering 1.2.0 


pavlenex:
https://docs.btcpayserver.org/Wallet/#signing-with-a-hot-wallet in 1.2.0 we've added an option that when a hot wallet is created it'll automatically use the seed store on a server to sign. We may want to update that part and also see if there are some updates regarding signing that are required.

dennisreimann:
Depending on how urgent this is, it might make sense to combine it with the upcoming changes in btcpayserver/btcpayserver#2713. Though they are related, this can be tackled on it's own -- just wanted to bring it up, as it would be an opportunity to rework the signing docs as a whole.

CypherSkunk:
I will give this one a go

pavlenex:
Sure, go ahead @CypherSkunk and let us know if you need any help.

petzsch:
@Bangalisch pointed me to this issue.

Can we use this issue to gather some ideas what needs to go into the documentation? I think we should (with the current version) mention that hot wallets might be needed for certain features like the On-chain Payout Processors? (need to check if that info is allready somewhere else)

I don't know did @CypherSkunk start writing something that we can flesh out? (ping)

I would like to open a PR for this one tomorrow  and start writing up something that I see fit.

Bangalisch:
can be closed @pavlenex

PR #1163, opened 2022-09-17
Slightly updated Wallet.md

Bangalisch:
Yoo yoo

Looked at this #950  one and found small things only. Also deleted a double entry for PSBT at the bottom where it's talked through in a section higher up. 

Maybe point should look at more of the Wallet page then just this.


petzsch (approved):
LGTM

Thank you for your contribution. 👍

pavlenex (approved):

PR #1167, opened 2022-09-24
Small updates on proper titles

Bangalisch:
Yoooo
just some small updates on proper titles on pages create wallet and testnet demo.

### DIFF
--- a/docs/Contribute/Write.md
+++ b/docs/Contribute/Write.md
@@ -2,7 +2,7 @@
 
 [[toc]]
 
-# Why documentation matters
+## Why documentation matters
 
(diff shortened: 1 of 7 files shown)


Bangalisch:
Is there a reason why the pages in the screenshot below are set to noindex somewhere? (didnt really search for where yet =X)

image

dennisreimann:
@Bangalisch Those are [pages that redirect](https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/.vuepress/redirects.js), because they have been moved. The general doc pages are getting indexed, but we don't want those redirect utility pages to be indexed.

dennisreimann (approved):
Thanks, good work!

PR #1179, opened 2022-10-30
Update wallet import

### LINKED ISSUE #1026: Improve `Connect an existing wallet > Import wallet file` and recommend it instead of xpub import (Update Electrum Doc, video)
I think we should probably update instructions and create a quick video of people importing Electrum into btcpay also configuring the gap limit as part of the onboarding, it's less stressful. 

Also the onboarding should explain the wallet file import which then configures master key and keypath in btcpay, where as xpub import doesn't. Will add an issue in docs for this.

Bangalisch:
Yo yo

Updated the Wallet import for electrum with a new video by @webwworthy. Replacing all the old gifs.

Also replaced 2 steps and their images to reflect the new UI.

closes #1026 unless @pavlenex also wants more textual changes in general. Gap limit does have an entry.

https://docs.btcpayserver.org/ElectrumWallet/#configuring-the-gap-limit-in-electrum


petzsch:
Can we still discuss about the order in which we present these options?

I as a user would still prefer the xpub over the wallet import. Not uploading my. private keys that could be stored in this electrum wallet file, sounds like a better idea.

Opinion on this? @pavlenex

pavlenex:
Wallet file is not your private key it's a public key with additional stuff needed , account key path, etc

pavlenex (approved):
Looks good. Why is a video linked unlisted @Bangalisch ?

PR #1189, opened 2022-11-17
Apps.md_PaymentButtonUpdates

raduButucelea23:
Updated the "Payment Button" instructions.

### DIFF
--- a/docs/Apps.md
+++ b/docs/Apps.md
@@ -61,7 +61,7 @@ If you would like to provide digital or physical products to the backers of your
 
 Easily-embeddable HTML and highly-customizable **payment buttons** allow users to receive tips and donations. Online stores can also integrate payment buttons. When a site visitor clicks on the button, BTCPay displays the **invoice**.
 
-1. In your left menu bar, find Payments and click Pay Button.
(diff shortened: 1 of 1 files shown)


docs/Apps.md
```diff
@@ -61,7 +61,7 @@ If you would like to provide digital or physical products to the backers of your
 
 Easily-embeddable HTML and highly-customizable **payment buttons** allow users to receive tips and donations. Online stores can also integrate payment buttons. When a site visitor clicks on the button, BTCPay displays the **invoice**.
 
-1. In your left menu bar, find Payments and click Pay Button.
+1. In your left menu bar, under the "PLUGINS" section, select "Manage Plugins" and select "Pay Button".
```
pavlenex, line 64: Thanks for spotting this inconsistency, however I don't think the suggested improvements are 100% correct. You don't have to click `Manage Plugins`, afair `Pay Button` is by default displayed in the sidebar under `Plugins`, can we rewrite it like that?

dennisreimann: That's correct, as a system plugin it is available without further configuration.

dennisreimann (approved):

PR #1194, opened 2022-11-25
Consistency  Nginx error/pr…

ndeet:
…oblem.

### DIFF
--- a/docs/Deployment/ReverseProxyToTor.md
+++ b/docs/Deployment/ReverseProxyToTor.md
@@ -126,7 +126,7 @@ Create a config file for the domain, e.g. `/etc/nginx/sites-available/btcpayserv
 ```nginx
 server {
   listen 80;
-  server_name mydomain.com;
+  server_name btcpayserver.mydomain.com;
 
   # Let's Encrypt verification requests
   location ^~ /.well-known/acme-challenge/ {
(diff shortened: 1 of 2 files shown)


docs/Deployment/ReverseProxyToTor.md
```diff
@@ -214,7 +214,18 @@ Restart nginx once more:
 systemctl restart nginx
 ```
 
-Now, visiting `mydomain.com` should show your BTCPay Server instance.
+Now, visiting `btcpayserver.mydomain.com` should show your BTCPay Server instance.
+
+:::tip
+If you see an Nginx error of "503 Service Temporarily Unavailable" or similar but your BTCPay Server is reachable otherwise then you need to make BTCPay Server aware of your new domain. You can do this by using environment variables (docker based setup), log into your BTCPay Server via SSH:
```
dennisreimann, line 220:
```suggestion
If you see an nginx error of "503 Service Temporarily Unavailable" or similar but your BTCPay Server is reachable otherwise, you need to make BTCPay Server aware of your new domain. You can do so by using environment variables (Docker based setup), log into your BTCPay Server via SSH:
```

ndeet:
Thanks for fixing the wording 🚀

PR #1216, opened 2023-01-05
Update WalletFAQ with mempoolfullrbf=1

Bangalisch:
Yo.

Just added short FAQ on " Does BTCPay Server use mempoolfullrbf=1 "


NicolasDorier:
ACK

PR #1226, opened 2023-02-07
PHP Greenfield API examples

ndeet:
- Added PHP GF examples
- Fixed some spelling according to Grammarly
- Fixed some spelling of GreenField -> Greenfield
- Deliberately named the PHP example filename "Greenfield" and not like the others "GreenField", will fix spelling through docs in a different PR.

Assumption: Afair we decided that Greenfield is spelled like this "Greenfield" and NOT "GreenField"


dennisreimann (approved):
utACK.

PR #1251, opened 2023-03-27
Add user documentation for Forms

NicolasDorier:
Once merged, we will need to add a link from BTCPay Server to this doc.


docs/Forms.md
```diff
@@ -0,0 +1,142 @@
...
+
+Output:
+
+![Form](./img/Forms-1.png "Form")
+
+In a field definition, only the following fields can be set:
+
+| Field   |      Description      |
+|----------|-------------|
+`.fields.constant` | If `true`, the `.value` must be set in the form definition, and the user will not be able to change the field's value. ( example: the form definition's version) |
```
petzsch, line 105:
```suggestion
|`.fields.constant` | If `true`, the `.value` must be set in the form definition, and the user will not be able to change the field's value. ( example: the form definition's version) |
```

Hopefully that fixes the problematic table formating in the [preview](https://deploy-preview-1251--btcpayserver-docs-preview.netlify.app/Forms/):
grafik

docs/Forms.md
```diff
@@ -0,0 +1,142 @@
...
+Some well-known names can be interpreted and modify the invoice's settings.
+
+| Field name   |      Description      |
+|----------|-------------|
+| `invoice_amount` | The invoice's amount |
+| `invoice_currency` | The invoice's currency |
+
+## Pre-filling form values
+
+You can pre-fill the fields of an invoice automatically by adding query strings to the form's URL, such as `?your_field=value`.
```
petzsch, line 134: Please clarify if pre-filling works for fields that have been set to constant: true

At least thats what stays as an open question for me when reading this. 😄

petzsch (changes requested):
Looks good to me, only 2 small remarks.

docs/Forms.md
dennisreimann: I don't think it does, but I can check if we can improve it with some CSS.

docs/Forms.md
NicolasDorier: @petzsch `constant: true` doesn't have any interaction with pre-filling.
Constant means that whatever its value set in the form, it will always be ignored. Instead, the value set in the template is always the one chosen.

Mandog1120 (commented):
`forms`

PR #1253, opened 2023-03-27
Add LNDhub API plugin docs

dennisreimann:
I will release the plugin together with the next BTCPay Server version.


ndeet (approved):
tested read trough it, works :+1:

PR #1258, opened 2023-04-01
Update Shopify.md

Bangalisch:
Added warning for Deprecate API calls linking back to original issue https://github.com/btcpayserver/btcpayserver/issues/4510

### DIFF
--- a/docs/Shopify.md
+++ b/docs/Shopify.md
@@ -11,7 +11,7 @@ The following document guides you through **setting up BTCPay Server with [Shopi
 [![BTCPay Server - Shopify Video](https://img.youtube.com/vi/jJjAyvgWVfk/mqdefault.jpg 'BTCPay Server - Shopify')](https://www.youtube.com/watch?v=jJjAyvgWVfk 'BTCPay Server - Shopify Video')
 
 :::warning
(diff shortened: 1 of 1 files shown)


docs/Shopify.md
```diff
@@ -10,6 +10,10 @@ The following document guides you through **setting up BTCPay Server with [Shopi
 
 [![BTCPay Server - Shopify Video](https://img.youtube.com/vi/jJjAyvgWVfk/mqdefault.jpg 'BTCPay Server - Shopify')](https://www.youtube.com/watch?v=jJjAyvgWVfk 'BTCPay Server - Shopify Video')
 
+:::warning
+Shopify might give false flags on deprecated API calls. The most likely answer we could find, due to a change by Shopify to specific fields of the `Order` resource and is assumed to get used in BTCPay Server. Find more detail [here](https://github.com/btcpayserver/btcpayserver/issues/4510)
```
petzsch, line 14:
```suggestion
Shopify might give false flags on deprecated API calls. The most likely answer we could find, due to a change by Shopify to specific fields of the `Order` resource and is assumed to get used in BTCPay Server. Find more details [here](https://github.com/btcpayserver/btcpayserver/issues/4510)
```

petzsch (changes requested):
LGTM.

personally I would say details instead of detail here. But it seems to be a case of personal [preference](https://english.stackexchange.com/questions/123672/detail-or-details-read-on-for-more-detail-details).

pavlenex (approved):

Issue #1264, opened 2023-04-14
YouTube Video Playback broken in many browsers


petzsch:
When trying to playback a youtube video in our docs, I get the following error:

![grafik](https://user-images.githubusercontent.com/1374810/231973482-cf679ace-5bd8-493d-962c-8d58f6f926af.png)

@pavlenex recommended bumping something, as sv2 also uses vuepress and their video playback still works.

Maybe @dennisreimann has an idea?

All Browsers tested with where affected:
- Chrome on Windows PC and Android
- Firefox on Windows Desktop and Android Mobile Version
- iOS Browsers also seem affected.

dennisreimann:
Thanks for reporting, fixed!

Bangalisch:
> Thanks for reporting, fixed!

Yooo

Just checked the solution and was like why take out all the alt text. I don't know if this is an actual solution. 
Searched further to find the actual moment this gone wrong and found a little thing with prettier 3 weeks ago. 
Running this changed all the "" into '' or from double to single quotes. The reason why the last video https://github.com/btcpayserver/btcpayserver-doc/pull/1255/files  worked was due to me doing them with the "". 

Just also tested this out with a PR (closed again by now to not have it merged). 

This was introduced due to using Prettier and this changed all the quotes..
https://github.com/btcpayserver/btcpayserver-doc/blame/ccf298a808b8564fce0d0e28e14198f7f30268ce/docs/Guide.md

dennisreimann:
@bas I fixed it already, but the alt text shouldn't appear there in the first place — so this was basically wrong all the time, but somehow it worked while using `"` instead of `'`. However, the solution is to completely remove the alt text from the URL part of the links.

ndeet:
@Bas02 @dennisreimann as discussed yesterday I added a markdown-it plugin that adjusts the parser to use the alt tags as title tags, see above linked PR. 

According to markdown spec the title needs to be in double quotes, so the prettier [commit](https://github.com/btcpayserver/btcpayserver-doc/commit/b1432054e147836d7286e1bae2f98e62f2752363) seems to have broken it as suggested by @Bas02 Maybe there is a way to avoid prettier to doing that, will also have a look at this.

ndeet:
[...]

PR #1265, opened 2023-04-14
Test PR for alt txt

Bangalisch:
This a Test PR to see solution of #1264 to be right or just a workaround.


Bangalisch:
Test seems to be correct.. 

Single qoutes where introduced three weeks ago by prettier usage.

PR #1271, opened 2023-04-24
New Infographic BTCPayServer Usecase.md

### LINKED ISSUE #1050: Add better visual for the `UseCase.md` and provide visual examples of people using BTCPay
Following up on #996 I think the infographic doesn't really convey information well, it looks odd and text-heavy. Instead of infographic, I suggest we add images of user-groups for paragraphs where this fits. This would include person using BTCPay point of sale app, online shop accepting BTCPay, freelancer requesting bill, etc. I think showcasing btcpay in a wild tells a better story than infographic.

webwworthy:
Updated use case flyer infographic.

closes #1050


Bas02:
When looking at the page Great job! Visual is much better than before. 
However, one comment, the QR goes towards Opensats as talked about, but there's no text above describing why. 

Please add according to our notes.

Bas02:
LGTM!

PR #1276, opened 2023-05-01
UseCase: Fix API link

### LINKED ISSUE #1274: Broken link in usecase.md
Just logging an issue someone reported, and I was able to replicate on iphone.

> Link pointing to greenfield api in usecase.md is broken when you click on the greenfield API, it only works after refresh

dennisreimann:
Closes #1274.


pavlenex (approved):
tack

PR #1279, opened 2023-05-04
Adding markdown-it plugin to parse alt attrs also as title.

ndeet:
I played a bit ping pong with ChatGPT and came up with a markdown-it plugin, to fix the title tag as discussed in dev call and on #1264 

- parses images and uses the found alt text and adds it as title
- you can still override the title according to mardown spec with double quotes like I did in the Guide.md as example `BTCPay Server`
- not sure about when prettier would change double quotes to single quotes, it works for me locally so it seems to be no problem as long as it keeps double quotes?

**Youtube**
For youtube videos if we would want to do that there too we would need to adjust the replaceYoutubeLinks in the preprocess here I think: https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/.vuepress/preprocessMarkdown.js#LL70C20-L70C20

It adds a link as wrapper to an iframe, the iframe title (which does not show up). If the alt attribute is available there we would need to adjust the regex to get it and set the title in the link tag to make it show up on mouseover.


dennisreimann (approved):
Added titles to the videos as well and updated the YT replacement script so that it works with either single or double quotes. The latter should be seen more as a nice to have and we should simply stick to using the standard Markdown syntax for the alt texts.

dennisreimann:
Yup, we can then also continue to run `prettier` every once in a while. I find that it helps to keep the formatting consistent.

PR #1282, opened 2023-05-16
Add kukks' plugins

dennisreimann:
Mr. Kukks went through the effort to create documentation for his plugins, we should display them more prominently.


Kukks (approved):
sexy, danke

PR #1318, opened 2023-07-31
Change c-lightning reference to CLN, minor fixes

ndeet:
- c-lightning to CLN
- remove Ptarmigan reference
- fix minor typos/wording


dennisreimann (approved):
Thatnks, looks good. Minor nitpick: the frequent "(CLN)" use seems a bit superfluous.

ndeet:
Yes I could reduce a few occassions and only leave it on the general pages and only mention it once on the LN FAQ but I thought it will be good for search and SEO so that stuff gets found when entered in the searchbar. Can change if you think it is too much.

PR #1336, opened 2023-09-12
Rewrite Accounting documentation to explain Reporting

NicolasDorier:
Follow up of https://github.com/btcpayserver/btcpayserver/pull/5293

The previous accounting export will be removed in the next version.
I was thinking of putting more screenshot and description here, but I believe lot's of changes to Reporting will happen in the short term, as such it's better to wait.


pavlenex (approved):
Concept ACK probably needs a lot more work as we add features.

PR #1379, opened 2024-03-22
Remove Shopware marketplace listing for now.

ndeet:
Shopware delisted the plugin in their marketplace because they wanted to charge a 2000,- EUR fee per year even it is an open source and free of charge extension. Crazy.


dennisreimann:
Shopware ngmi.

dennisreimann (approved):

PR #1385, opened 2024-04-06
[MERGE AFTER 2.0] Remove docs on customizing checkout page

TChukwuleta:
Based on the [BTCPay Server 2.0: Megathread](https://github.com/btcpayserver/btcpayserver/issues/5899#top) , part of the task is to Remove docs on https://docs.btcpayserver.org/Development/Theme/#checkout-page-themes

This PR tends to resolve it, by removing the path from the side bar so that it is not accessible


NicolasDorier:
Why not removing the article itself and the images?

TChukwuleta:
I wasn't sure so sure, that's why I didn't remove it. I would do that.


Yea, I'm aware that it's after 2.0 merge. Thank you

dennisreimann (changes requested):
I dont think we should completely remove the page, but most of it. We should keep the "Developing and extending a custom theme" part and adapt it to what we've added with the branding options. 

Can you re-add that page with the "Developing and extending a custom theme" only (until "2. Bootstrap themes") and I can adjust it from there? Thanks!

TChukwuleta:
Sure @dennisreimann

TChukwuleta:
Updated @dennisreimann

dennisreimann (approved):
Adapted it to fit with the current state.

pavlenex (approved):

pavlenex:
**To be merged when 2.0 is out.**

pavlenex:
@dennisreimann mentioned 

> The doc in the PR version is a refresher about customizing the server theme, imho it can get merged right away.

I agree with this, @NicolasDorier ?

NicolasDorier:
Seems good to me

PR #1387, opened 2024-04-23
Restructure API examples, add more

### LINKED ISSUE #1371: API examples streamlining
Add more consistent examples through curl, php, js

ecommerce use case (accept payments):
- create invoice
- register webhook?
- process webhook
- issue refund

more sophisticated, management examples
- create store
- create user
- create api key

ndeet:
The idea is outlined in issue #1371 where we structure examples for eCommerce use case and also on managing BTCPay Server by creating users and stores. All the examples for PHP, Node.js and cURL have the same structure while differing in details according to language, e.g. you can't really do a webhook validation endpoint with cURL.

- [x] cURL
- [x] PHP
- [x] NodeJS
- [x] overview
- [x] testing

Fixes #1371

### DIFF
--- a/docs/Development/GreenFieldExample-NodeJS.md
+++ b/docs/Development/GreenFieldExample-NodeJS.md
@@ -1,35 +1,57 @@
-# Greenfield API example with Node.JS (JavaScript)
+# Greenfield API example with Node.js (JavaScript)
+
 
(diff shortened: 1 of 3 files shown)


pavlenex:
Great, a dev would need to review these to get merged though.

dennisreimann (approved):
ACK. If I understand btcpayserver#5822 correctly, we refrain from adding the examples inline with the API docs and I think this is fine with these updates.

PR #1397, opened 2024-05-05
Fix minimum PHP version.

ndeet:


pavlenex:
Why is it a draft, seems pretty straightforward fix?

PR #1396, opened 2024-05-04
shopify docs improvement

TChukwuleta:
Updates shopify docs (#1352)

Issue #1352, opened 2023-12-21
Revise and improve branding, readability and SEO for integration pages.


pavlenex:
**Context:**
Often, new users discover BTCPay Server by finding a way to accept bitcoin in their e-commerce software (CMS). This is usually done by Googling "Accept bitcoin/crypto in XYZ". 

It seems with a bit of cleanups and improvements, that allows for better experience overal, we can boost discoverability of those pages.

When a user lands on a page Shopify, by googling How to accept Bitcoin in Shopify for example ,they're expecting to get context, understand the features/advantages and how to get started as quickly as possible.

This issue is a tracker for multiple smaller issues that we can tackle, by striving to improve our integrations landing pages. In most cases the readme of the integration repo has to be addressed since we pull docs from them. Please reach out before tackling any of this issues for consensus and approach alignment. 

Good example on how a page should look is our newly updated WooCommerce plugin.

Ideal description of an integration should have:
1. How to accept Bitcoin in [Integration name]
2. Quick introduction on [Integration name] + BTCPay Server
3. Short description of how BTCPay works linking to relevant document
4. Highlight features concisely and relevant for specific target group - online merchants
5. Have visually pleasing graphics that outline purpose of the integration and showcase important features merchants want
6.Have clear instructions on how to connect integration to BTCPay Server 
7. Have a video tutorial on how to do integration
8. Provide links on how to get more information and support
9. [Optional]Explaining advantages/disadvantages or competitor analysis graph

Sorted in order of priority, each of these should be tackled:

```[tasklist]
### Tasks
- [ ] WooCommerce add graphics + screenshots 
- [ ] Shopify revise 
- [ ] Magento 2
- [ ] PrestaShop
- [ ] Joomla/Virtumart
- [ ] OpenCart
- [ ] Drupal
- [ ] Zapier
- [ ] Shopaware
- [ ] SmartStore
- [ ] NopCommerce [TBD]
- [ ] BigCommerce [TBD]
```

pavlenex:
It's still a work in progress, I'd expect @ndeet to work on completing it after BTCPrague.

TChukwuleta:
Hey @pavlenex  @ndeet not sure I have encountered Shopaware before.

If this exists please can you point me to a guide or anything

ndeet:
@TChukwuleta there was a typo, it is Shopware https://docs.btcpayserver.org/Shopware/#plugin-for-shopware-6

[...]

PR #1399, opened 2024-05-07
Shopify docs cleanup

ndeet:
- Fix steps numbering; 
- Fix steps, add more info; 
- Remove duplicate img/Shopify directory and files.

Related to #1352 and PR #1396 

cc @pavlenex


ndeet:
Not sure why image 11 and 12 don't show up, seems all other images are cached and have some string attached but not those two new files. I would assume it will work and is just a quirk in netlify.

pavlenex:
@ndeet LGMT for images perhaps folder lower/uppercase difference?
Screenshot 2024-05-08 at 17 37 06

ndeet:
@pavlenex what I did was deleted "Shopify" with the two 11 + 12 images and then did new screenshots and put them into "shopify" folder with a new file extension of .png. The images show up locally so I think it is fine. Maybe you can pull and run npm run start to double check?

ndeet:
@pavlenex found the problem, forgot to but ./ to make image path relative, fixed now also in preview.

pavlenex (approved):

PR #1402, opened 2024-05-18
Update payment requests and remove form builder

TChukwuleta:
@pavlenex reupdated payment request content and images, also took out the form builder and the advanced form builders content and navigation path


pavlenex (approved):
LGTM

PR #1415, opened 2024-07-09
Adding odoo to docs.

ndeet:


pavlenex:
LGTM, But I assume our goal is to update those docs? Should we merge it once updated?

ndeet:
I updated them and created separate one for the 17.0 branch as the process is slightly different, before I created this PR.

ndeet:
Which means all ready to go and announce.

pavlenex:
So what should we do, here's what I see:
Screenshot 2024-07-15 at 18 24 12
Should we default to 17 with a link to 16, this doc seems quite oudated in terms of graphics at least 😆

pavlenex:
Here you go @ndeet , not perfect but will do the job for now.

Odoo-16

Odoo-17

pavlenex:
@ndeet there's a conflict now FYI.

PR #1420, opened 2024-08-15
Update ManualDeployment.md

FADM-DCMN-CORY-A-HOFSTAD-USN:
The installation works better on Ubuntu 24 with the backports repository.


dennisreimann:
What does "works better" mean here? I think we should update that doc to follow the [MS installation guidelines for Ubuntu 24](https://learn.microsoft.com/en-us/dotnet/core/install/linux-ubuntu-install?tabs=dotnet8&pivots=os-linux-ubuntu-2410#ubuntu-2410).

dennisreimann:
Please reopen with more details if this is still relevant.

PR #1429, opened 2024-09-25
correct redundant buffer call causing error, add typescript workaround

bnbaltz:
The first edit was to fix Buffer.from being called again, Buffer.from does not accept a string argument and outputs bytes. The current code had the following flow: 
x = Buffer.from(digest<string>)
y = Buffer.from(x<bytes>) -> Error, expecting str

It was simple to just remove y as it's redundant. 

The second edit was to allow for those using TypeScript to overcome the simple errors they will run into.


dennisreimann (approved):
Thanks for the contribution!

PR #1468, opened 2024-11-29
Remove Connecting Additional BTCPay Services

### LINKED ISSUE #1455: Remove `Connecting Additional BTCPay Services` with `Plugins`
This section is quite outdated and should be removed since neither transmuter or librepatron are maintained anylonger 

https://docs.btcpayserver.org/WhatsNext/#connecting-additional-btcpay-services

instead we could add plugins and users could extend functionality of BTCPay once they play around with core features.

TChukwuleta:
Resolves #1455 

Cc. @pavlenex

### DIFF
--- a/docs/WhatsNext.md
+++ b/docs/WhatsNext.md
@@ -36,7 +36,7 @@ e.t.c
 
 ## Extending BTCPay Server: Plugins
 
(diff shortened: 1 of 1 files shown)


pavlenex (approved):
Great improvement Tobses!

PR #1502, opened 2025-04-15
include more faq

TChukwuleta:
More Shopify FAQ based on user feedback

### DIFF
--- a/docs/ShopifyV2.md
+++ b/docs/ShopifyV2.md
@@ -210,7 +210,7 @@ When you click on it you can see the payment details:
 
 ## FAQ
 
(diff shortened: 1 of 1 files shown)


docs/ShopifyV2.md
```diff
@@ -210,6 +210,8 @@ When you click on it you can see the payment details:
 
 ## FAQ
 
+- Can I use ShopifyV1 and ShopifyV2 together? No, ShopifyV1 and ShopifyV2 should not be used together. Using both versions simultaneously can lead to unexpected behavior, such as duplicate order record on Shopify. We recommend that you disable Shopify V1 and stick to only ShopifyV2
```
pavlenex, line 213: LGTM, Let's try to be consistent in naming. 
```suggestion
- Can I use Shopify V1 and Shopify V2 together? No, Shopify V1 and Shopify V2 should not be used together. Using both versions simultaneously can lead to unexpected behavior, such as duplicate order record on Shopify. We recommend that you disable Shopify V1 and stick to only Shopify V2
```

pavlenex (approved):

PR #1527, opened 2025-09-16
Update shopify v2 docs with latest shopify migration flow

TChukwuleta:
PR updates documents to cater for new UI flow from Shopify


ndeet (approved):
Thank you :raised_hands:

PR #1536, opened 2025-11-26
Docs for Easy Digital Downloads

ndeet:
Also updated GiveWP docs with some missing headline and support info.


pavlenex:
should we embed @webwworthy's video now before we merge it, on top of the page? @ndeet

PR #1541, opened 2025-12-16
Add Subscriptions.md and Monetization.md

pavlenex:
This PR adds Subscription.md in preparation for the upcoming 2.3.0 release. It uses PR description that @NicolasDorier provided and all the videos he made as a starting point. I leveraged AI for this one, but it was done in a responsible way, meaning data set was strictly controlled and I've done at least 2 rounds of reviews and deployment locally before shipping it. I've tested it locally to ensure everything works, but would appreciate reviews here for sanity check.

Since we're time constrained, I'd appreciate concrete review suggestions instead of comments, so we can move forward in the most productive way possible.

### DIFF
--- a/docs/Monetization.md
+++ b/docs/Monetization.md
@@ -1,6 +1,6 @@
 # Monetization
 
-Monetization lets a BTCPay Server administrator charge users of their BTCPay Server instance using the built in [Subscriptions](http://docs.btcpayserver.org/Subscriptions) system. It is intended for operators who host BTCPay Server for others and want to charge for server access, maintenance, or technical support.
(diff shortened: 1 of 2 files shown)


docs/Monetization.md
```diff
@@ -0,0 +1,63 @@
+# Monetization
+
+Monetization lets a BTCPay Server administrator charge users of their BTCPay Server instance using the built in [Subscriptions](http://docs.btcpayserver.org/Subscriptions) system. It is intended for operators who host BTCPay Server for others and want to charge for server access, maintenance, or technical support.
```
ndeet: remove `,` after "maintenance"

docs/Subscriptions.md
```diff
@@ -0,0 +1,262 @@
+# Subscriptions
+
+Subscriptions allow you to manage recurring payments for services you provide, such as newsletters, SaaS products, memberships, or recurring donations.
```
ndeet: remove `,` after "memberships"

PR #1545, opened 2026-01-21
Update shopify v2 docs with refunds flow

TChukwuleta:
include refund flow for Shopify plugin

@pavlenex  @ndeet

### DIFF
--- a/docs/ShopifyV2.md
+++ b/docs/ShopifyV2.md
@@ -216,10 +216,12 @@ When you click on it you can see the payment details:
 
 ## Shopify Refunds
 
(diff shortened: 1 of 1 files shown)


NicolasDorier:
@TChukwuleta 

```
In the webhook settings, click `Create webhook`, and in the modal select the following:
- For event, select Refund create
- Format should be JSON
- URL is the webhook URL you copied from Shopify refund settings on BTCPay Server page
- Webhook API version choose 2026-01
```

Please can you look if we can remove this step?
I would be surprised if you can't achieve this through the `toml` template file from the deployer project.

NicolasDorier:
@TChukwuleta ok it makes sense

About this screen `Go to Settings -> Notifications -> webhook`...

image


I think you can simplify this by just make a link that the user just have to click on?

pavlenex (approved):
LGTM

pavlenex:
@ndeet @NicolasDorier please give it a final look and then we can merge it.

PR #1579, opened 2026-04-02
Fix hamburger menu not displaying on homepage

### LINKED ISSUE #1578: Hamburger menu is not displaying on index page
image

Rest of the pages are fine
image

Task is to show menu on the homepage of docs in same way its called for the other pages.

For example image

tmchow:
## Summary

The mobile hamburger menu was missing from the homepage (index page) while showing correctly on all other pages. The custom `Homepage.vue` layout didn't include the `Navbar` and `Sidebar` components that the default VuePress theme layout provides.

## Changes

Added the `Navbar` and `Sidebar` components to `Homepage.vue`, wrapped the content in the standard `theme-container` div, and wired up the sidebar toggle state. This matches how the default VuePress `Layout.vue` handles navigation, so the hamburger menu now appears on mobile just like it does on every other page.

## Testing

Verified the component imports resolve correctly against the existing custom `Navbar.vue` and `Sidebar.vue` in the theme. The sidebar is initialized with an empty items array since the homepage doesn't need sidebar navigation, only the toggle button (hamburger menu).

Fixes #1578

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/docs/.vuepress/theme/layouts/Homepage.vue
+++ b/docs/.vuepress/theme/layouts/Homepage.vue
@@ -16,11 +16,6 @@
     aria-labelledby="main-title"
   >
     <header class="hero">
-      <img
-        v-if="data.heroImage"
-        :src="$withBase(data.heroImage)"
-        :alt="data.heroAlt || $title"
-      >
       <button type="button" class="btcpay-theme-switch" @click="toggleColorMode($event)">
         <svg viewBox="0 0 10 10" xmlns="http://www.w3.org/2000/svg">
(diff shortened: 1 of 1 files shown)


pavlenex:
Looks off on desktop.

Screenshot 2026-04-04 at 11 07 10

We can also probably remove that icon from the hero altogether on both mobile and desktop. Looks odd that logo is duplicated in menu header and then in hero.

Screenshot 2026-04-04 at 11 07 52

tmchow:
Good call on both points. Pushed a fix (5e896ba) that removes the hero image since the Navbar already shows the logo - no need for it in both places.

The desktop layout issue should also be resolved now that the hero section no longer has the large logo competing with the navbar. Let me know if it still looks off after the deploy preview updates.

pavlenex (approved):

PR #1601, opened 2026-07-06
Remove Voltage from docs.

ndeet:
context https://x.com/voltage_cloud/status/2072721707194786042


.gitignore
```diff
@@ -1,4 +1,5 @@
 /.vscode
+/.claude
```
pavlenex, line 2: Why this?

pavlenex:
@ndeet can you resolve conflict please?

.gitignore
ndeet: because it contains only local claude specific config and gets created when you run claude; adding it to .gitignore makes sure your local config is not committed to the repo, maybe we should also add .codex which is same but for codex

ndeet:
> @ndeet can you resolve conflict please?

Sorry was not checking my gmail for some time. Done.

